PR #1, opened 2025-02-20
Refactor LLM interface into a separate package

justinsb:
We bring in the llm abstraction from Config Connector (a cut down version),
and we trivially invoke it.


janetkuo (approved):
/lgtm

gollm/vertexai.go
```diff
@@ -0,0 +1,318 @@
...
+	if err != nil {
+		return nil, fmt.Errorf("finding default credentials: %w", err)
+	}
+	opts = append(opts, option.WithCredentials(creds))
+
+	projectID := ""
+	location := "us-central1"
+
+	if projectID == "" {
+		cmd := exec.CommandContext(ctx, "gcloud", "config", "get", "project")
```
mikebz, line 50: Curious about your thoughts on taking a dependency on `gcloud` (even an optional one).

If the design of the plug in to interface with multiple models via their API's it might be best to keep the interfaces as a web service call

go.mod
```diff
@@ -0,0 +1,44 @@
+module github.com/GoogleCloudPlatform/kubectl-ai
+
+go 1.23.2
+
+replace github.com/GoogleCloudPlatform/kubectl-ai/gollm => ./gollm
+
+require github.com/GoogleCloudPlatform/kubectl-ai/gollm v0.0.0-00010101000000-000000000000
```
mikebz, line 7: curious why this is needed and if it is it might be good to leave a comment for posterity.

gollm/vertexai.go
justinsb: So the "current project" is only a gcloud concept, there's no GCP API call for this.

agent.go
```diff
@@ -179,7 +180,18 @@ func (a *Agent) AskLLM(ctx context.Context) (*ReActResponse, error) {
 
 	logger.Info("Thinking...", "prompt", prompt)
 
-	return a.ContentGenerator.GenerateContent(ctx, a.Model, prompt)
+	response, err := a.ContentGenerator.GenerateCompletion(ctx, &gollm.CompletionRequest{
```
droot, line 183: like the name :)

gollm/gemini.go
```diff
@@ -0,0 +1,368 @@
...
+
+	geminiParts = append(geminiParts, genai.Text(request.Prompt))
+
+	log.Info("sending GenerateContent request to gemini", "parts", geminiParts)
+	geminiResponse, err := model.GenerateContent(ctx, geminiParts...)
+	if err != nil {
+		return nil, err
+	}
+
+	if len(geminiResponse.Candidates) > 1 {
```
droot, line 107: nice!

gollm/gemini.go
```diff
@@ -0,0 +1,368 @@
...
+		default:
+			return nil, fmt.Errorf("unexpected type of content part: %T", part)
+		}
+	}
+
+	return &GeminiCompletionResponse{geminiResponse: geminiResponse, text: response.String()}, nil
+}
+
+// StartChat starts a new chat with the model.
+func (c *GeminiClient) StartChat(systemPrompt string) Chat {
```
droot, line 132: sweet! I will try this out this week.

droot (approved):
This is awesome! Thank you.

PR #2, opened 2025-02-21
initial import of kubernetes AI agent

droot:
The purpose of this PR is to show current state of AI agent implementation and seek guidance on code structure.

I am also open to importing this code as-is and refactor as we go along. To keep things simple, I have removed other LLM implementations and `json` mode. I do have those in my local code if needed.


main.go
```diff
@@ -0,0 +1,229 @@
...
+			return // Exit if client initialization fails
+		}
+		defer client.Close()
+
+		contentGenerator = &GeminiLLM{
+			Client: client,
+			Model:  *model,
+		}
+	case "ollama":
+		// plug in ollama
```
janetkuo, line 92: Should we remove this case for now given that it doesn't do anything?

main.go
```diff
@@ -0,0 +1,229 @@
...
+// models
+var geminiModels = []string{
+	"gemini-2.0-flash-thinking-exp-01-21",
+}
+
+func main() {
+	// non interactive execution when query is specified on the command line.
+	queryFromCmd := flag.String("query", "", "query for the agent")
+	maxIterations := flag.Int("max-iterations", 20, "maximum number of iterations")
+	kubeconfig := flag.String("kubeconfig", "/Users/sunilarora/.kube/auto-ai-1", "path to the kubeconfig file")
```
janetkuo, line 26: choose a different default?

main.go
```diff
@@ -0,0 +1,229 @@
...
+	}
+	defer logFile.Close()
+
+	textHandler := slog.NewTextHandler(logFile, nil)
+	logger := slog.New(textHandler)
+
+	mdRenderer, err := glamour.NewTermRenderer(
+		glamour.WithAutoStyle(),
+		glamour.WithPreservedNewLines(),
+		// glamour.WithWordWrap(2000),
```
janetkuo, line 48: nit: is this line needed?

main.go
droot: Done.

main.go
droot: Defaulted to empty value. This will lead to default `kubectl` behavior.

main.go
droot: Removed it.

droot:
Thanks @janetkuo for the review, addressed all the three comments.

agent.go
```diff
@@ -0,0 +1,346 @@
+package main
+
+import (
+	"context"
+	"fmt"
+	"os"
+	"strings"
+	"text/template"
+
+	"github.com/charmbracelet/glamour"
```
justinsb, line 10: We'll have to get a license aggregator going for release

agent.go
```diff
@@ -0,0 +1,346 @@
...
+	}
+
+	// Handle max iterations reached
+	logger.Info("Max iterations reached", "iterations", a.CurrentIteration)
+	fmt.Printf("\nSorry, Couldn't complete the task after %d attempts.\n", a.MaxIterations)
+	return fmt.Errorf("max iterations reached")
+}
+
+// executeAction handles the execution of a single action
+func (a *Agent) executeAction(ctx context.Context, action *Action, workDir string) (string, error) {
```
justinsb, line 130: We should unify this with "normal" function calling

agent.go
```diff
@@ -0,0 +1,346 @@
...
+	}
+	if role != "system" {
+		a.Messages = append(a.Messages, msg)
+	}
+	if a.tracePath == "" {
+		return nil
+	}
+	if err := WriteToFile(a.tracePath, fmt.Sprintf("%s: %s\n", role, content)); err != nil {
+		logger.Error("Error writing to trace file", "error", err)
+		// tracing is secondary to the main logic, so ignore the failures.
```
justinsb, line 204: We might want to pull out the tracing into its own package, or just use logging / structured logging / spans

agent.go
```diff
@@ -0,0 +1,346 @@
...
+	}
+	return nil
+}
+
+func (a *Agent) History() string {
+	var history strings.Builder
+	for _, msg := range a.Messages {
+		history.WriteString(fmt.Sprintf("%s: %s\n", msg.Role, msg.Content))
```
justinsb, line 212: It feels like this is duplicating the Chat API that we get from Gemini, and that we should probably push it down into the LLM layer (e.g. from #1).  Different models will have different optimal ways of formatting the sequence of messages (based on what they were trained with)

agent.go
```diff
@@ -0,0 +1,346 @@
...
+
+type Message struct {
+	Role    string `json:"role"`
+	Content string `json:"content"`
+}
+
+// WriteToFile appends the given content to a file.
+// If the file does not exist, it is created.
+// Returns an error if any operation fails.
+func WriteToFile(fileName string, content string) error {
```
justinsb, line 237: Nit: we should call this AppendToFile (and probably just refactor to a tracing layer)

agent.go
```diff
@@ -0,0 +1,346 @@
...
+	Tools       string
+}
+
+// generateFromTemplate generates a string from the ReAct template using the provided data.
+func (a *Agent) generateFromTemplate(data Data) (string, error) {
+	var tmpl *template.Template
+	var err error
+
+	if a.templateFile != "" {
+		// Read custom template file
```
justinsb, line 268: Idea: probably going to be easier to use `// go:embed`

llm.go
```diff
@@ -0,0 +1,97 @@
...
+			if err == iterator.Done {
+				return modelNames, nil
+			}
+			return nil, err
+		}
+		modelNames = append(modelNames, strings.TrimPrefix(m.Name, "models/"))
+	}
+}
+
+// parseReActResponse parses the LLM response into a ReActResponse struct
```
justinsb, line 63: I like this code, I think we should move llm.go into a package (I called it gollm in #1), and then some LLMs will "natively" support function calling or structured output, and some will do better when we parse JSON as here.  So 👍  as a starting point!

main.go
```diff
@@ -0,0 +1,226 @@
...
+	Queries []string `json:"queries"`
+	Model   string   `json:"model"`
+}
+
+func (s *session) PreviousQueries() string {
+	return strings.Join(s.Queries, "\n")
+}
+
+// Logger key for context
+type loggerKey struct{}
```
justinsb, line 198: Cool.  We may be able to use `klog.FromContext`, but I think we should merge this PR and then move this into its own package.

justinsb (approved):

PR #4, opened 2025-02-25
Added k8s-bench

droot:
Added `k8s-bench`, an evaluation benchmark to assess performance of LLM models for real-world kubernetes tasks. Eval tasks are located under `k8s-bench/tasks/`.

For now, I am using this as an e2e test for `kubectl-ai` agent.

```sh

# build the k8s-bench binary
go build

# evaluate model performance for scale related tasks
./k8s-bench --agent-bin <path/to/kubectl-ai/binary> --task-pattern scale --kubeconfig <path/to/kubeconfig>
```

Running the benchmark will produce results as below:

```sh
Evaluation Results:
==================

Task: scale-deployment
  Provider: gemini
    gemini-2.0-flash-thinking-exp-01-21: true

Task: scale-down-deployment
  Provider: gemini
    gemini-2.0-flash-thinking-exp-01-21: true

```

### DIFF
--- a/k8s-bench/eval.go
+++ b/k8s-bench/eval.go
@@ -1,0 +1,162 @@
+package main
+
+import (
+	"fmt"
+	"os"
+	"os/exec"
+	"path/filepath"
+	"strings"
+
+	"gopkg.in/yaml.v2"
+)
+
+func runEvaluation(config EvalConfig) error {
+	tasks, err := loadTasks(config)
+	if err != nil {
+		return fmt.Errorf("failed to load tasks: %w", err)
+	}
+
+	results := make(map[string]map[string]map[string]bool) // task -> provider -> model -> success
+
+	for taskID, task := range tasks {
+		fmt.Printf("Evaluating task: %s\n", taskID)
+
+		// Run setup if specified
+		if task.Setup != "" {
+			setupPath := filepath.Join(config.TasksDir, taskID, task.Setup)
+			if err := runScript(setupPath, config.KubeConfig); err != nil {
+				return fmt.Errorf("setup failed for task %s: %w", taskID, err)
+			}
+		}
+
+		for _, provider := range config.LLMProviders {
+			results[taskID] = make(map[string]map[string]bool)
+			results[taskID][provider] = make(map[string]bool)
+
+			for _, model := range config.Models[provider] {
+				success, err := evaluateTask(config, taskID, task, provider, model)
+				if err != nil {
+					fmt.Printf("Error evaluating task %s with %s/%s: %v\n", taskID, provider, model, err)
+					results[taskID][provider][model] = false
+					continue
+				}
+				results[taskID][provider][model] = success
+			}
+		}
+	}
+
+	printResults(results)
+	return nil
+}
+
+func loadTasks(config EvalConfig) (map[string]Task, error) {
+	tasks := make(map[string]Task)
+
+	entries, err := os.ReadDir(config.TasksDir)
+	if err != nil {
+		return nil, err
+	}
+
+	for _, entry := range entries {
+		if !entry.IsDir() {
+			continue
+		}
+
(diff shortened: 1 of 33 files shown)


k8s-bench/eval.go
```diff
@@ -0,0 +1,162 @@
...
+		}
+
+		tasks[taskID] = task
+	}
+
+	return tasks, nil
+}
+
+func evaluateTask(config EvalConfig, taskID string, task Task, provider, model string) (bool, error) {
+	// Run kuba
```
janetkuo, line 95: nit: why is it called kuba

k8s-bench/eval.go
```diff
@@ -0,0 +1,162 @@
...
+	"gopkg.in/yaml.v2"
+)
+
+func runEvaluation(config EvalConfig) error {
+	tasks, err := loadTasks(config)
+	if err != nil {
+		return fmt.Errorf("failed to load tasks: %w", err)
+	}
+
+	results := make(map[string]map[string]map[string]bool) // task -> provider -> model -> success
```
justinsb, line 19: We might want to introduce a struct here, but ... OK :-)

k8s-bench/eval.go
```diff
@@ -0,0 +1,162 @@
+package main
+
+import (
+	"fmt"
+	"os"
+	"os/exec"
+	"path/filepath"
+	"strings"
+
+	"gopkg.in/yaml.v2"
```
justinsb, line 10: Nit: we should use sigs.k8s.io/yaml if we can

k8s-bench/eval.go
```diff
@@ -0,0 +1,162 @@
...
+	cmd.Env = append(os.Environ(), fmt.Sprintf("KUBECONFIG=%s", config.KubeConfig))
+	cmd.Stdout = os.Stdout
+	cmd.Stderr = os.Stderr
+	fmt.Printf("\nRunning kuba for task %s with %s/%s\n", taskID, provider, model)
+
+	if err := cmd.Run(); err != nil {
+		return false, fmt.Errorf("kuba failed: %w", err)
+	}
+
+	success := true
```
justinsb, line 112: Nit: we might consider a string result, if there's no verifier I'm wary of calling it a "success" ?

k8s-bench/tasks/configure-ingress/setup.sh
```diff
@@ -0,0 +1,15 @@
...
+# Create namespace and deploy a test application
+kubectl create namespace ingress-test
+
+# Create deployment and service
+kubectl create deployment web-app --image=nginx -n ingress-test
+kubectl expose deployment web-app --name=web-service --port=80 -n ingress-test
+
+# Wait for deployment to be ready
+for i in {1..30}; do
+    if kubectl get deployment web-app -n ingress-test -o jsonpath='{.status.availableReplicas}' | grep -q "1"; then
```
justinsb, line 11: Nit: kubectl wait ?

justinsb: Idea: we should ask the LLM to translate this to kubectl wait 😂   "How do I use kubectl to wait for a deployment to be ready?"

k8s-bench/tasks/configure-ingress/task.yaml
```diff
@@ -0,0 +1,6 @@
+goal: "Please create route for routing traffic from path '/app' to service 'web-service' in namespace 'ingress-test'."
+setup: "setup.sh"
```
justinsb, line 2: Idea: default these and use the presence of scripts to decide whether to run them

justinsb:
A few nits, but nothing blocking I don't think...

/approve
/lgtm

k8s-bench/eval.go
janetkuo: +1, maybe just throw an error when there's no verifier

PR #5, opened 2025-02-26
k8s-bench: start to write structured results

justinsb:
- **eval: create helper script to run evals**
- **k8sbench: switch to sigs.k8s.io/yaml**
- **k8s-bench: start to write structured results**

### DIFF
--- a/dev/ci/periodics/run-evals
+++ b/dev/ci/periodics/run-evals
@@ -10,6 +10,12 @@ REPO_ROOT="$(git rev-parse --show-toplevel)"
 cd ${REPO_ROOT}
 
 
+if [[ -z "${OUTPUT_DIR:-}" ]]; then
+    OUTPUT_DIR="${REPO_ROOT}/.build/k8sbench"
+    mkdir -p "${OUTPUT_DIR}"
+fi
+echo "Writing results to ${OUTPUT_DIR}"
+
 BINDIR="${REPO_ROOT}/.build/bin"
 mkdir -p "${BINDIR}"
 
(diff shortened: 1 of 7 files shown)


justinsb:
/assign @droot

These evals are awesome ... I'm wondering if we can run them periodically!  But at least we can start by capturing the output in a structured way.

dev/ci/periodics/run-evals
```diff
@@ -0,0 +1,30 @@
...
+BINDIR="${REPO_ROOT}/.build/bin"
+mkdir -p "${BINDIR}"
+
+cd "${REPO_ROOT}"
+go build -o "${BINDIR}/kubectl-ai" .
+
+cd "${REPO_ROOT}/k8s-bench"
+go build -o "${BINDIR}/k8s-bench" .
+
+# evaluate model performance for scale related tasks
```
droot, line 28: nit: comment is not relevant because the following command runs all tasks.

k8s-bench/eval.go
```diff
@@ -7,7 +7,8 @@ import (
 	"path/filepath"
 	"strings"
 
-	"gopkg.in/yaml.v2"
+	"github.com/GoogleCloudPlatform/kubectl-ai/k8s-bench/pkg/model"
+	"sigs.k8s.io/yaml"
```
droot, line 11: thank you :)

droot (approved):

droot:
> These evals are awesome ... I'm wondering if we can run them periodically! But at least we can start by capturing the output in a structured way.

Thank you :). We have two dependencies:
    1. k8s cluster: `kind` helps here. I have used kind cluster using `podman` on my mac successfully.
    2. LLM: we can use a mock LLM here (will be closer to unit testing) or we can use `ollama` with gemma or similar model. For `gemma` model we will treat it to succeed if some percentage threshold success rate it met. We can also use real Gemini model if the frequency is not that high.

dev/ci/periodics/run-evals
justinsb: Doh, thanks - will fix next time I touch this file :-)

PR #6, opened 2025-02-26
Fix typo in prompt

justinsb:
Remarkably, these don't seem to actually confuse the LLM!

### DIFF
--- a/agent.go
+++ b/agent.go
@@ -306,7 +306,7 @@ func (a *Agent) generateFromTemplate(data Data) (string, error) {
 }
 
 // Move the default template to a constant
-const defaultTemplate = `You are a Kubernetes Assitant tasked with answering the following query:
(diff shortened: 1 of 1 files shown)


agent.go
```diff
@@ -306,7 +306,7 @@ func (a *Agent) generateFromTemplate(data Data) (string, error) {
 }
 
 // Move the default template to a constant
-const defaultTemplate = `You are a Kubernetes Assitant tasked with answering the following query:
+const defaultTemplate = `You are a Kubernetes Assistant tasked with answering the following query:
```
droot, line 309: Good catch!

One time I changed `assistant` to `expert`, the agent stopped solving the tasks instead started giving too much *advise* :)

It will be fun to play with prompts in future :)

droot (approved):

PR #11, opened 2025-02-26
Made "query" a positional argument and updated README.md

droot:
Major changes:

1. Made `query` argument positional to enhance the overall UX. 

    Now users can simply do:

    ```sh
    kubectl-ai "double the capacity of nginx app"
    ```
    
    instead of
    
    ```sh
    
    kubectl-ai --query "double the capacity of nginx app"
    ```

2. Improved the README with installation/usage. Added a screenshot (show vs tell).
3. Defaulted `--kubeconfig` commandline arg to environment variable or `~/.kube/config`.


main.go
```diff
@@ -72,6 +72,29 @@ func run(ctx context.Context) error {
...
+			}
+			kubeconfigPath = filepath.Join(homeDir, ".kube", "config")
+		}
+	}
+
+	// Check for positional arguments (after all flags are parsed)
+	args := flag.Args()
+	var queryFromCmd string
+	if len(args) > 0 {
+		// Use the first positional argument as the query
```
justinsb, line 94: We might want to want if there are more positional args, as this will be a common failure mode (forgetting to quote)

justinsb (approved):

README.md
```diff
@@ -1,7 +1,77 @@
...
+
+![screenshot of kubectl-ai](kubectl-ai.png)
+
+## Installation
+
+### Prerequisites
+
+- Go 1.23 or later
+- kubectl installed and configured
+- Access to an Gemini API (other supported model)
```
janetkuo, line 15:
```suggestion
- Access to a Gemini API key (or other supported AI models)
```

README.md
```diff
@@ -1,7 +1,77 @@
...
+# Build and install
+go install .
+```
+
+### Environment Setup
+
+Set up Gemini API key as follows:
+
+```bash
+export GEMINI_API_KEY=your_api_key_here
```
janetkuo, line 33: We might need to rename the env to be more generic for other models

main.go
janetkuo: +1, so perhaps throwing an error here is better

justinsb: I meant "We might want to check ..."

README.md
droot: I have seen LLM provider prefix as a common practice in the tools that support multiple LLM providers e.g. ANTHROPIC_API_KEY and such. If ever, agent makes use of multiple providers simultaneously for example reasoner/thinking model for planning and regular model for tool use, having distinct keys will be useful. So keeping it as-is.

main.go
droot: Added check. Also added support for taking input from STDIN so that one can use `kubectl-ai` in pipe syntax for composite commands.

For example:

```sh
cat query.txt | kubectl-ai -
```

```sh
# not very useful example :)
echo "check health for nginx app in hello namespace" | kubectl-ai

README.md
droot: Done.

PR #12, opened 2025-02-28
Added support for direct stdin support

droot:
Removed the "-" argument for reading content from STDIN. It makes `kubectl-ai` consistent with how unix commands handle `stdin`.

I was surprised how simple it was to add.


droot:
@janetkuo @justinsb I am merging this in and we can always revert if this turn out be a bad idea.

PR #15, opened 2025-03-06
Refactor terminal rendering into its own package

justinsb:


justinsb:
I think we want to make both rendering and the react strategy pluggable.  I tried doing both at once and that was a pretty drastic change (#15).  I think splitting them up should make it much more reviewable, so this is just the UI change

droot (approved):
Thanks for cleaning up my mess :)

PR #16, opened 2025-03-06
Refactor: start to split out strategy

justinsb:
We want to compare multiple LLM strategies.

Start by refactoring in the same package, we will then rename.

### DIFF
--- a/agent.go
+++ b/agent.go
@@ -11,28 +11,46 @@ import (
 
 	"github.com/GoogleCloudPlatform/kubectl-ai/gollm"
 	"github.com/GoogleCloudPlatform/kubectl-ai/pkg/journal"
+	"github.com/GoogleCloudPlatform/kubectl-ai/pkg/llmstrategy"
 	"github.com/GoogleCloudPlatform/kubectl-ai/pkg/ui"
 	"k8s.io/klog/v2"
 )
 
 // Agent knows how to execute a multi-step task. Goal is provided in the query argument.
 type Agent struct {
-	AgentType          AgentType
-	Query              string
-	Model              string
-	PastQueries        string
-	LLM                gollm.Client
-	Messages           []Message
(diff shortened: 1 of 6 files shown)


justinsb:
Splitting the big one up into reviewable PRs @droot .  The good news is that I also have a simple function-execution strategy teed up, so this won't be too bad!

agent.go
```diff
@@ -11,27 +11,44 @@ import (
 
 	"github.com/GoogleCloudPlatform/kubectl-ai/gollm"
 	"github.com/GoogleCloudPlatform/kubectl-ai/pkg/journal"
+	"github.com/GoogleCloudPlatform/kubectl-ai/pkg/llmstrategy"
 	"github.com/GoogleCloudPlatform/kubectl-ai/pkg/ui"
 	"k8s.io/klog/v2"
 )
 
 // Agent knows how to execute a multi-step task. Goal is provided in the query argument.
 type Agent struct {
```
droot, line 20: another option to organize will be have different agent type structs, that all use common pieces to implement an agent type and top level `SuperAgent` is composed of sub-agents and in this case, only subagent type (ReAgent or ChatBasedAgent).

agent.go
```diff
@@ -11,27 +11,44 @@ import (
...
-	Query            string
-	Model            string
-	PastQueries      string
+	Model string
+
+	Recorder journal.Recorder
+
+	Strategy llmstrategy.Strategy
```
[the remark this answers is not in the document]
justinsb: I do think we might have non-agentic strategies.  Is deep-thinking still agentic?  Is RAG?

agent.go
justinsb: I have no objection to refactoring the commonalities once we have the two types :-)

droot (approved):

PR #17, opened 2025-03-06
Refactor: move strategies into their own packages

justinsb:
- **Refactor: start to split out strategy**
- **refactor: move ReAct strategy into its own package**


justinsb:
Builds on #16

PR #26, opened 2025-03-07
ci: add tests for formatting and autogen

justinsb:


justinsb:
(Hoping that this fails verify-format, as I will then push the header fixes)

droot (approved):

PR #31, opened 2025-03-09
fix: don't use kubectl run --timeout

justinsb:
I saw a message saying that the --timeout flag on kubectl run is ignored,
so switching to use bash timeout instead.


justinsb:
The message I saw was:

```
Running verifier for task fix-service-routing
Flag --timeout has been deprecated, because it is not used by this command. It will be removed in version 1.29.
If you don't see a command prompt, try pressing enter.
```

droot (approved):

PR #29, opened 2025-03-08
Ollama support

droot:
This builds on top of https://github.com/GoogleCloudPlatform/kubectl-ai/pull/28 so pl. do not merge this one.

PR #32, opened 2025-03-10
Added confirmation step (configurable) when resources are modified 

droot:
Though we want to make agent fully autonomous, but it is still scary when agent modifies resources in the cluster.

This change adds support for confirming with the user whenever if it is about to run a command that modifies resources in the cluster.

An example run:

```sh
>> bump up the capacity of nginx app by 50%
  Running: kubectl get deployments
  To list all deployments in the current namespace and identify the deployment
  related to 'nginx app'.

  Running: kubectl get deployments --all-namespaces
  To list all deployments across all namespaces and find the deployment related to
  'nginx app'.

  Running: kubectl get deployment nginx -n hello -o yaml
  To get the current configuration of the nginx deployment, including the number
  of replicas.

  Running: kubectl scale deployment nginx --replicas=3 -n hello
  To scale the 'nginx' deployment in the 'hello' namespace to 3 replicas,
  increasing its capacity by 50%.

  Are you sure you want to run this command (Y/n)?  <------here is how confirmation will show up
Y

  Successfully bumped up the capacity of the nginx app by 50%. The nginx
  deployment in the 'hello' namespace is now scaled to 3 replicas.
```

Note: we are trusting LLM to asses if a kubectl command will modify the resources or not, but an LLM could be wrong or go rogue. I think here another LLM can be used to confirm (like peer reviewing a change :)).


pkg/llmstrategy/chatbased/strategy.go
```diff
@@ -134,8 +143,16 @@ assistant: kubectl get pod my-pod -o jsonpath='{.status.phase}'
...
 					functionName := call.Name
 					command, _ := call.Arguments["command"].(string)
-					log.Info("function call", "functionName", functionName, "command", command)
+					modifies_resource, _ := call.Arguments["modifies_resource"].(string)
+					log.Info("function call", "functionName", functionName, "command", command, "modifies_resource", modifies_resource)
 					u.RenderOutput(ctx, fmt.Sprintf("  Running: %s\n", command), ui.Foreground(ui.ColorGreen))
+					if a.AsksForConfirmation && modifies_resource == "yes" {
+						confirm := u.AskForConfirmation(ctx, "  Are you sure you want to run this command (Y/n)? ")
+						if !confirm {
+							u.RenderOutput(ctx, "Sure.\n", ui.RenderMarkdown())
```
justinsb, line 152: Idea: I wonder if we can reply with some sort of error that tells the LLM that the user said not to run this command...

pkg/llmstrategy/react/strategy.go
```diff
@@ -318,7 +329,8 @@ If you need to use a tool:
     "action": {
         "name": "Tool name (kubectl, gcrane, cat, echo)",
         "reason": "Explanation of why you chose this tool (not more than 100 words)",
-        "input": "complete command to be executed."
+        "input": "complete command to be executed.",
+		"modifies_resource": "Whether the command modifies a kubernetes resource. Possible values are 'yes' or 'no' or 'unknown'"
```
justinsb, line 333: Nit: hopefully this fails our go format check (?)

justinsb (approved):

justinsb:
Only reviewed second commit, as I assume the first one is #29 

/approve
/lgtm

PR #41, opened 2025-03-11
gollm: add a simple scheme for env-var based LLM configuration

justinsb:
Starting with just the scheme e.g.

* LLM_CLIENT=vertexai://
* LLM_CLIENT=gemini://
* LLM_CLIENT=ollama://

### DIFF
--- a/gollm/factory.go
+++ b/gollm/factory.go
@@ -1,0 +1,45 @@
+// Copyright 2025 Google LLC
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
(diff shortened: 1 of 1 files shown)


gollm/factory.go
```diff
@@ -0,0 +1,45 @@
...
+
+import (
+	"context"
+	"fmt"
+	"net/url"
+	"os"
+)
+
+// NewClient builds an Client based on the LLM_CLIENT env var
+func NewClient(ctx context.Context) (Client, error) {
```
droot, line 25: I have been thinking, for `model`, adopting syntax to specify both llm-provider and llm-model.

for example:

`gemini/gemma3b`, `ollama/gemma3b`, `gemini/gemini-2.0-flash`

WDYT ?

droot (approved):

gollm/factory.go
justinsb: 💯  this is just a starting point, and I also think it's fine if you want to break compatibility at this stage.

e.g. `ollama://10.0.0.1/gemma3b:8b?contextWindow=16384`

PR #42, opened 2025-03-12
kubectl runner handles input command correctly

droot:
Major bug fix.

kubectl runner was splitting the incoming command into an argument array before passing to `os.Exec` and that was causing commands with multiple lines to be dropped out. This was eliminating very useful commands that involves `here-doc` syntax and `kubectl apply -f -`.

During the POC, I rushed into intercepting `kubectl edit` command and incorrectly handled it. This fix should unlock major performance gains. I am already seeing gemma3 performing super well.

### DIFF
--- a/pkg/llmstrategy/react/react_prompt_template_default.txt
+++ b/pkg/llmstrategy/react/react_prompt_template_default.txt
@@ -22,8 +22,8 @@ If you need to use a tool:
     "thought": "Your detailed reasoning about what to do next",
     "action": {
         "name": "Tool name ({{.Tools}})",
-        "reason": "Explanation of why you chose this tool (not more than 100 words)",
-        "input": "complete command to be executed.",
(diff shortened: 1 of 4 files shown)


justinsb (approved):

pkg/llmstrategy/react/react_prompt_template_default.txt
```diff
@@ -22,8 +22,8 @@ If you need to use a tool:
     "thought": "Your detailed reasoning about what to do next",
     "action": {
         "name": "Tool name ({{.Tools}})",
-        "reason": "Explanation of why you chose this tool (not more than 100 words)",
-        "input": "complete command to be executed.",
+        "reason": "Explanation of why you chosse this tool (not more than 100 words)",
```
janetkuo, line 25: choose?

droot: Aah... thanks for catching this. fixing it.

PR #46, opened 2025-03-13
evals: added create resources evals (pod and networkpolicy)

droot:
Some more evals for creating resources.


droot:
Merging this PR in because I am generating eval summary and need this in. We can always tweat these in follow up PRs.

PR #51, opened 2025-03-18
Add eval that checks output

justinsb:
- **evals: able to check output**
- **evals: add eval to list images for pods**


droot (approved):
This is great. Will help write the trouble-shooting evals because most of them are about UI output vs doing something to the resources.

PR #62, opened 2025-03-20
Add trivy tool

justinsb:
- **Basic configuration in ~/.config/kubectl-ai/config.yaml**
- **Use a conversational logical model**
- **Have functions return typed output**
- **tools: add simple tool to call trivy**


pkg/llmstrategy/chatbased/strategy.go
```diff
@@ -39,84 +41,116 @@ type Strategy struct {
...
 
 	Kubeconfig          string
 	AsksForConfirmation bool
 
 	Tools tools.Tools
 }
 
-// ExecuteChatBased executes a chat-based agentic loop with the LLM using function calling.
-func (a *Strategy) RunOnce(ctx context.Context, query string, previousQueries []string, u ui.UI) error {
+type Conversation struct {
```
droot, line 58: Good name.

pkg/llmstrategy/chatbased/strategy.go
```diff
@@ -39,84 +41,116 @@ type Strategy struct {
...
+	strategy *Strategy
```
droot, line 59: this name is somehow not that intuitive but we will live with it for now.

In my mind, I think of this as a the agentic loop, so sort of a `looper` or `runner`. May be we can rename this to `AgentRunner` or something.

pkg/llmstrategy/chatbased/strategy.go
```diff
@@ -39,84 +41,116 @@ type Strategy struct {
...
+	// Recorder captures events for diagnostics
+	recorder journal.Recorder
+
+	UI      ui.UI
+	llmChat gollm.Chat
+
+	workDir string
+}
+
+func (s *Strategy) NewConversation(ctx context.Context, u ui.UI) (llmstrategy.Conversation, error) {
```
droot, line 70: I am bit confused about the hierarchy here. conversation has strategy and then strategy has a method to start a conversation .....

droot (approved):

pkg/llmstrategy/chatbased/strategy.go
justinsb: Yeah, it could certainly use some improvement.  Conversation is a stateful discussion thread.  Strategy is the pluggable way we interact with the conversation.  We could have two conversations going on both with the same react strategy.

justinsb: Oh I see what you mean.  Yes, it turns out we need some variables from the "parent strategy".  Ideally we would not.  I managed to extract that in our eval framework by copying the variables.  We could see if that works here.  And maybe `parent` is a better name.

pkg/llmstrategy/chatbased/strategy.go
justinsb: To clarify (because I got confused), you mean a better name than Strategy? (vs Conversation)  I agree, Strategy is not a great name ... it is too abstract :-)

droot: I was referring to `Strategy` not being a good name.

pkg/llmstrategy/interfaces.go
```diff
@@ -16,10 +16,22 @@ package llmstrategy
...
+	NewConversation(ctx context.Context, userInterface ui.UI) (Conversation, error)
+}
+
+type Conversation interface {
+	// Close should be called to free up resources
+	io.Closer
+
+	// RunOneRound will send the query to the LLM, and go through cycles with the LLM,
+	// evaluating requested functions until we reach a stopping point.
+	RunOneRound(ctx context.Context, query string) error
```
droot, line 36: Something doesn't feel right about these interfaces.

So my mental model is:

I want to have a conversation with a particular style (ReAct or ChatBased), so one way to represent will be:

```

conversationStyle = ReAct{} or ChatBased{}
_ = conversation.Do(style ReActOrChatBased)

```

droot (commented):
I think the implementation is correct, but I am still not sure about the abstractions.

pkg/llmstrategy/chatbased/strategy.go
justinsb: Agreed and have no objection to trying any better names you come up with :-)

justinsb:
Merged in another PR

PR #64, opened 2025-03-20
Start to support multi-step conversations

justinsb:
- **k8s-bench: change syntax to support multiple-step interactions**
- **k8s-bench: start to support multi-step eval**

### DIFF
--- a/k8s-bench/main.go
+++ b/k8s-bench/main.go
@@ -30,7 +30,6 @@ import (
 )
 
 type Task struct {
-	Goal       string `json:"goal"`
 	Setup      string `json:"setup,omitempty"`
 	Verifier   string `json:"verifier,omitempty"`
 	Cleanup    string `json:"cleanup,omitempty"`
@@ -38,6 +37,12 @@ type Task struct {
 	Disabled   bool   `json:"disabled,omitempty"`
 
 	Expect []Expectation `json:"expect,omitempty"`
+
+	Script []ScriptStep `json:"script,omitempty"`
+}
+
+type ScriptStep struct {
+	Prompt string `json:"prompt"`
 }
 
 type Expectation struct {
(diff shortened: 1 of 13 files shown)


janetkuo (approved):

k8s-bench/main.go
```diff
@@ -30,14 +30,19 @@ import (
...
-	Goal       string `json:"goal"`
 	Setup      string `json:"setup,omitempty"`
 	Verifier   string `json:"verifier,omitempty"`
 	Cleanup    string `json:"cleanup,omitempty"`
 	Difficulty string `json:"difficulty"`
 	Disabled   bool   `json:"disabled,omitempty"`
 
 	Expect []Expectation `json:"expect,omitempty"`
+
+	Script []ScriptStep `json:"script,omitempty"`
```
droot, line 41: Not sure what's happening here. Does `Script` here means `ChatMessages` or `ChatScript` ?

k8s-bench/main.go
```diff
@@ -30,14 +30,19 @@ import (
...
+}
+
+type ScriptStep struct {
+	Prompt string `json:"prompt"`
```
droot, line 45: hmm...is this the `user message` or prompt that we sent to LLM ?

droot (commented):
I think part of refactoring is good but not so sure about the `script/prompt` input.

k8s-bench/main.go
justinsb: Script is the prompts we send.  It replaces `goal`, and allows us to have multiple prompts

k8s-bench/main.go
justinsb: The prompt we send.  We move `goal` to `script.prompt`, and follow the steps of the `script`

justinsb: (Happy to do whatever name you can come up with, I liked the idea of following a script!)

droot (approved):

PR #65, opened 2025-03-20
Support cluster isolation

justinsb:
- **k8s-bench: change syntax to support multiple-step interactions**
- **k8s-bench: start to support multi-step eval**
- **k8s-bench: allow running each task in its own cluster, easily**

### DIFF
--- a/k8s-bench/main.go
+++ b/k8s-bench/main.go
@@ -39,7 +39,18 @@ type Task struct {
 	Expect []Expectation `json:"expect,omitempty"`
 
 	Script []ScriptStep `json:"script,omitempty"`
+
+	// Isolation can be set to automatically create an isolated cluster
+	// TODO: support namespaces also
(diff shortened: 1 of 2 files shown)


k8s-bench/main.go
```diff
@@ -30,14 +30,30 @@ import (
...
 	Difficulty string `json:"difficulty"`
 	Disabled   bool   `json:"disabled,omitempty"`
 
 	Expect []Expectation `json:"expect,omitempty"`
+
+	Script []ScriptStep `json:"script,omitempty"`
+
+	// Isolation can be set to automatically create an isolated cluster
+	// TODO: support namespaces also
+	Isolation IsolationMode `json:"isolation,omitempty"`
```
droot, line 45: I think we are mixing two things here: task definition with the environment against which tasks are run.

droot (approved):

PR #69, opened 2025-03-21
Removed Strategies to have one agentic loop

droot:
Note to reviewers: recommend looking at individual commints to track how we ended up in the final state. I am yet to update `k8s-bench` and run evals on it.

TL;DR

- Removed different strategies and now we have one chat-based agentic conversation
- ReAct is now implemented as a tool-use-shim to support models that don't support tool-use/function-calling natively. (We may move this to gollm but I don't think that's priority)
- Lots of changes to simplify the structure.


pkg/llmstrategy/react/strategy.go
```diff
@@ -56,41 +56,45 @@ type Strategy struct {
...
 	}
 
 	log.Info("Created temporary working directory", "workDir", workDir)
+	data := PromptData{
+		Tools: s.Tools,
+	}
+
+	systemPrompt, err := s.generatePrompt(ctx, defaultSystemPromptTemplate, data)
+	if err != nil {
+		log.Error(err, "Failed to generate system prompt")
```
justinsb: Nit: Why this pattern instead of just `return fmt.Errorf("failed to generate system prompt: %w", err)`

pkg/llmstrategy/chatbased/chatbased_systemprompt_template_default.txt
```diff
@@ -1,8 +1,40 @@
 You are a Kubernetes Assistant and your role is to assist a user with their kubernetes related queries and tasks. Your goal is to reason about the query and decide on the best course of action to answer it accurately.
 
+{{if .EnableToolUseShim }}
```
justinsb: Will be interesting to A/B test this, I think I've seen some prompts that give a description of the tools even though the LLM model natively supports tools (I.e. has the function definitions).  Another pattern is to give instructions or one-shot style examples to the LLM of when to use a tool.

pkg/llmstrategy/chatbased/strategy.go
```diff
@@ -269,15 +344,87 @@ func (a *Strategy) generatePrompt(_ context.Context, defaultPromptTemplate strin
...
+
+// PromptData represents the structure of the data to be filled into the template.
+type PromptData struct {
+	Query string
+	Tools tools.Tools
+
+	EnableToolUseShim bool
+}
+
+func (a *PromptData) ToolsAsJSON() string {
```
justinsb: Nice use of functions here 👍

main.go
```diff
@@ -41,6 +40,7 @@ import (
 
 // models
 var geminiModels = []string{
+	"gemini-2.0-flash",
 	"gemini-2.0-flash-thinking-exp-01-21",
```
justinsb, line 44: Does gemini-2.0-flash-thinking-exp-01-21 support structured output?

main.go
```diff
@@ -63,25 +63,29 @@ func main() {
...
-	Strategy   string `json:"strategy,omitempty"`
 	ProviderID string `json:"llmProvider,omitempty"`
 	ModelID    string `json:"model,omitempty"`
 
 	// AsksForConfirmation is a flag to ask for confirmation before executing kubectl commands
 	// that modifies resources in the cluster.
 	AsksForConfirmation bool `json:"askForConfirmation,omitempty"`
 
+	// EnableToolUseShim is a flag to enable tool use shim.
+	// TODO(droot): figure out a better way to discover if the model supports tool use
```
justinsb, line 74: Agreed, but actually I might want to evaluate gemini-flash with built-in functions vs our functions :-)

main.go
```diff
@@ -64,25 +63,29 @@ func main() {
...
 
 func (o *Options) InitDefaults() {
 	// default to react because default model doesn't support function calling.
-	o.Strategy = "react"
 	o.ProviderID = "gemini"
 	o.ModelID = geminiModels[0]
 	// default to false because our goal is to make the agent truly autonomous by default
 	o.AsksForConfirmation = false
 	o.MCPServer = false
+	o.EnableToolUseShim = false
```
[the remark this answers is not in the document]
justinsb: Good question.  I've been thinking about the idea of "profiles", like there are good combos of LLM provider, model, prompt, tool-use configuration etc.  Or at least some combos make a lot more sense than others 😂

pkg/llmstrategy/chatbased/strategy.go
```diff
@@ -428,3 +374,66 @@ func toMap(v any) (map[string]any, error) {
 	}
 	return m, nil
 }
+
+func candidateToShimCandidate(candidate gollm.Candidate) (*ShimCandidate, error) {
+	for _, part := range candidate.Parts() {
+		if text, ok := part.AsText(); ok {
+			parsedReActResp, err := parseReActResponse(text)
```
justinsb: Clever!

justinsb (approved):

justinsb:
Sorry for comment spamming ... looks good to me - some thoughts but nothing that should be addressed in this PR, just avenues for further exploration.

I think this is a great simplification so +100 from me :-)

PR #70, opened 2025-03-22
Better configuration loading

justinsb:
The config directory on MacOS is not ~/.config and is somewhat surprising to me.

Add a multi-path loader and also try loading from ~/.config.

### DIFF
--- a/main.go
+++ b/main.go
@@ -98,18 +98,32 @@ func (o *Options) LoadConfiguration(b []byte) error {
 	return nil
 }
 
-func run(ctx context.Context) error {
-	// Command line flags
-	var opt Options
-	opt.InitDefaults()
+func (o *Options) LoadConfigurationFile() error {
+	configPaths := []string{
+		"{CONFIG}/kubectl-ai/config.yaml",
+		"{HOME}/.config/kubectl-ai/config.yaml",
+	}
 
-	{
+	for _, configPath := range configPaths {
 		// Try to load configuration
-		configDir, err := os.UserConfigDir()
-		if err != nil {
-			return fmt.Errorf("getting user config directory: %w", err)
+		tokens := strings.Split(configPath, "/")
+		for i, token := range tokens {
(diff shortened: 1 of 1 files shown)


main.go
```diff
@@ -98,18 +98,32 @@ func (o *Options) LoadConfiguration(b []byte) error {
 	return nil
 }
 
-func run(ctx context.Context) error {
-	// Command line flags
-	var opt Options
-	opt.InitDefaults()
+func (o *Options) LoadConfigurationFile() error {
+	configPaths := []string{
+		"{CONFIG}/kubectl-ai/config.yaml",
```
droot, line 103: should we standardize on `${CONFIG}/.kubectl-ai` (hidden `kubectl-ai` ) ?

main.go
```diff
@@ -118,11 +132,25 @@ func run(ctx context.Context) error {
 				fmt.Fprintf(os.Stderr, "warning: could not load defaults from %q: %v\n", configPath, err)
 			}
 		}
-		if err := opt.LoadConfiguration(configBytes); err != nil {
-			fmt.Fprintf(os.Stderr, "warning: error loading configuration from %q: %v\n", configPath, err)
+		if len(configBytes) > 0 {
```
droot, line 135: does this mean we have a bug currently for users who don't have config file (like me) wrt to defaults ?

droot (approved):

main.go
droot: It will be nice to share a sample config in the PR description. Can serve as documentation.

main.go
janetkuo: We'd still run `opt.InitDefaults()` when there's no config files

janetkuo (approved):

main.go
droot: So, `opt.InitDefaults()` is called before this loading step and I was wondering if the content read is nil, will it reset the passed object. I tested, it doesn't, so no bugs.

PR #72, opened 2025-03-23
WIP: Switched to official genai sdk

droot:
Something I noticed `AI Studio` docs suggest using another Go SDK when I was troubleshooting an error for `gemma-3-27b-it` model, API was just returing 400 error without explanation.

Then I looked into AI studio developers docs and found out the official recommended SDK for Go is different one from the one we are using. The SDK we are using has not been updated since early Dec (last year). The official SDK is periodically maintained.

In this `work in progress` PR, I am trying to switch to new SDK and my first impression has been positive.

First, I was able to get the exact error for the `gemma` model, it's about not allowing system prompt `https://discuss.ai.google.dev/t/gemma-3-missing-features-despite-announcement/71692`. I have a workaround to test that as well.

side note: This surfaced because we switched to `chat-based` model. The previous model never used system prompt.

This has been quite some ride.

DO NOT MERGE because a few of the methods (list-models etc.) have not been migrated yet and I need to clean up a few things.


droot:
@justinsb @janetkuo the PR is good for final review.

gollm/gemini.go
```diff
@@ -67,24 +66,18 @@ var _ Client = &GeminiClient{}
...
+	for model, err := range c.client.Models.All(ctx) {
 		if err != nil {
-			if err == iterator.Done {
-				break
-			}
-			return nil, err
+			return nil, fmt.Errorf("error listing models: %w", err)
 		}
-		modelNames = append(modelNames, strings.TrimPrefix(m.Name, "models/"))
+		modelNames = append(modelNames, strings.TrimPrefix(model.Name, "models/"))
```
justinsb, line 73: Might be interesting to see what other metadata we have here (re our discussion to figure out whether models support function calling!)

gollm/gemini.go
```diff
@@ -114,111 +107,98 @@ func (c *GeminiClient) SetResponseSchema(responseSchema *Schema) error {
...
+		},
+		history: []*genai.Content{},
 	}
 
-	if systemPrompt != "" {
-		model.SystemInstruction = &genai.Content{
-			Parts: []genai.Part{
-				genai.Text(systemPrompt),
-			},
+	if c.model == "gemma-3-27b-it" {
```
justinsb, line 158: Interesting .... TBD whether this belongs here or in the caller.  But let's start with this and see how we feel over time.

justinsb (approved):

gollm/gemini.go
droot: Good pt. Quick look https://pkg.go.dev/google.golang.org/genai#Model shows `SupportedActions` as something interesting. Will look into it.

```
// List of actions that are supported by the model.
	SupportedActions [][string](https://pkg.go.dev/builtin#string) `json:"supportedActions,omitempty"`
```

gollm/gemini.go
droot: Ack.

pkg/agent/systemprompt_template_default.txt
```diff
@@ -1,4 +1,4 @@
-You are a Kubernetes Assistant and your role is to assist a user with their kubernetes related queries and tasks. Your goal is to reason about the query and decide on the best course of action to answer it accurately.
+You are a `kubectl-ai`, an AI assistant with expertise in operating and performing actions against a kubernetes cluster. Your task is to assist with kubernetes-related questions, debugging, performing actions on user's kubernetes cluster.
```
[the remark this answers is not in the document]
droot: Good catch. Fixed.

gollm/gemini.go
```diff
@@ -114,111 +107,98 @@ func (c *GeminiClient) SetResponseSchema(responseSchema *Schema) error {
...
+		// Note: gemma-3-27b-it does not allow system prompt
+		// xref: https://discuss.ai.google.dev/t/gemma-3-missing-features-despite-announcement/71692
+		// TODO: remove this hack once gemma-3-27b-it supports system prompt
+		chat.genConfig.SystemInstruction = nil
+		chat.history = []*genai.Content{
+			{Role: "user", Parts: []*genai.Part{{Text: systemPrompt}}},
```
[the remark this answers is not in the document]
droot: Ack. Will do in one of the upcoming PRs.

Note that this is only applicable if using `gemma-3` with google API, not with ollama etc. (actually this approach is now same as ollama client).

PR #73, opened 2025-03-25
modelserving: add cuda support

justinsb:


droot (approved):

modelserving/k8s/llm-server.yaml
```diff
@@ -15,8 +15,21 @@ spec:
       serviceAccountName: llm-server
       containers:
       - name: llm-server
-        image: fake.registry/llm-image:latest
-
+        image: gcr.io/gcp-project-id/llamacpp-gemma3-12b-it:latest
```
janetkuo, line 18: nit: Add a comment to update gcp-project-id

janetkuo (approved):

modelserving/k8s/llm-server.yaml
justinsb: Good call, I added a comment (I also updated this PR pretty substantially)

justinsb:
Cleaned this up so it's much more reasonable, added selective image building

modelserving/dev/tasks/build-images
```diff
@@ -17,5 +17,92 @@ if [[ -z "${TAG:-}" ]]; then
...
+
+  docker buildx build ${BUILDX_ARGS} \
+    -f images/llamacpp-server/Dockerfile \
+    --target rpc-server \
+    -t ${IMAGE_PREFIX}rpc-server-cuda:${TAG} \
+    --build-arg BASE_IMAGE=nvidia/cuda:12.6.2-runtime-ubuntu24.04 \
+    --build-arg BUILDER_IMAGE=nvidia/cuda:12.6.2-devel-ubuntu24.04 \
+    --build-arg "CMAKE_ARGS=-DGGML_RPC=ON -DGGML_CUDA=ON -DCMAKE_CUDA_ARCHITECTURES=all -DCMAKE_EXE_LINKER_FLAGS=-Wl,--allow-shlib-undefined" \
+    --progress=plain .
+fi
```
droot, line 108: I am surprised that the `llama.cpp` project doesn't have these images available somewhere we can copy from in the `deploy` step.

droot (approved):

modelserving/dev/tasks/build-images
justinsb: They have some images: https://github.com/ggml-org/llama.cpp/blob/master/docs/docker.md

But I want to add the model.

That said, I probably should have at least looked at how they are building their images, might have saved me some time 😂

PR #75, opened 2025-03-26
modelserving: add distributed serving

justinsb:


nicslatts:
IIUC this PR is to serve models in the cluster where kubectl-ai is running, and run them on CPUs or GPUs via CUDA (there seems to be both CPU and CUDA install modes), as opposed to calling out to a hosted model like Gemini. Is that accurate?

janetkuo:
> IIUC this PR is to serve models in the cluster where kubectl-ai is running, and run them on CPUs or GPUs (there seems to be both CPU and CUDA install modes), as opposed to calling out to a hosted model like Gemini. Is that accurate?

@nicslatts yes. I'm adding a `README` to this folder; hope it's helpful: https://github.com/GoogleCloudPlatform/kubectl-ai/pull/106

justinsb:
I've been thinking about this for a while now.  I think we should probably get this in, get the next one in (efficient model pulling), and then also add vLLM-backed serving.  vLLM and llama.cpp are both great and once we can support two I will feel more comfortable that our serving stack is "open" (i.e. can support more implementations)

droot (approved):

PR #76, opened 2025-03-26
Cleaned up the code

droot:
- main.go was too cluttered for my taste, so cleaned it up (always keep the entropy in check :))
 - removed the support for updating the model in the interactive shell because it requires restarting the conversation and just wasn't worth the complexity.
 - removed unnecessary model fetching. Fetch only when needed and cache it
 - subcommands such as `version`, `models` now works for non-interactive session as well.

### DIFF
--- a/gollm/factory.go
+++ b/gollm/factory.go
@@ -21,18 +21,21 @@ import (
 	"os"
 )
 
-// NewClient builds an Client based on the LLM_CLIENT env var
-func NewClient(ctx context.Context) (Client, error) {
-	s := os.Getenv("LLM_CLIENT")
-	if s == "" {
-		return nil, fmt.Errorf("LLM_CLIENT is not set")
-	}
-	u, err := url.Parse(s)
-	if err != nil {
-		return nil, fmt.Errorf("parsing LLM_CLIENT URL: %w", err)
+// NewClient builds an Client based on the LLM_CLIENT env var or the provided providerID. ProviderID (if not empty) overrides the provider from LLM_CLIENT env var.
+func NewClient(ctx context.Context, providerID string) (Client, error) {
+	if providerID == "" {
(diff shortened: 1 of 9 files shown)


gollm/factory.go
```diff
@@ -21,18 +21,21 @@ import (
...
-func NewClient(ctx context.Context) (Client, error) {
-	s := os.Getenv("LLM_CLIENT")
-	if s == "" {
-		return nil, fmt.Errorf("LLM_CLIENT is not set")
-	}
-	u, err := url.Parse(s)
-	if err != nil {
-		return nil, fmt.Errorf("parsing LLM_CLIENT URL: %w", err)
+// NewClient builds an Client based on the LLM_CLIENT env var or the provided providerID. ProviderID (if not empty) overrides the provider from LLM_CLIENT env var.
+func NewClient(ctx context.Context, providerID string) (Client, error) {
```
justinsb, line 25: Cool - I imagine one day we will extend the LLM_CLIENT "uri" to be more than just a providerID, but an obvious extension here if/when we do!

gollm/gemini.go
```diff
@@ -130,15 +122,15 @@ func (c *GeminiClient) GenerateCompletion(ctx context.Context, request *Completi
 }
 
 // StartChat starts a new chat with the model.
-func (c *GeminiClient) StartChat(systemPrompt string) Chat {
+func (c *GeminiClient) StartChat(systemPrompt string, model string) Chat {
```
justinsb, line 125: So I was thinking that the model might go into the LLM_CLIENT URI, but let's see how this feels from a caller perspective!  I hadn't anticipated the idea of switching models mid-chat when thinking about a URI.

justinsb (approved):
Nice refactor!

PR #79, opened 2025-03-26
Fixed bug in retrying logic

droot:
Realized I had a bug in the retry logic in the way we detect if error is retryable. 

There are more improvements I want to but will keep it for later PRs. For example (take the feedback from the API about backoff delay) and some minor improvements for other providers. Overall, this is working pretty nicely especially today got lots of 429 errors for `gemini-2.5-pro-experimental`.

### DIFF
--- a/gollm/factory.go
+++ b/gollm/factory.go
@@ -75,6 +75,8 @@ func (e *APIError) Unwrap() error {
 }
 
 // IsRetryableFunc defines the signature for functions that check if an error is retryable.
+// TODO (droot): Adjust the signature to allow underlying client to relay the backoff
+// delay etc. for example, Gemini's error codes contain retryDelay information.
 type IsRetryableFunc func(error) bool
 
 // DefaultIsRetryableError provides a default implementation based on common HTTP codes and network errors.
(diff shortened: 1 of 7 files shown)


gollm/factory.go
```diff
@@ -152,18 +154,18 @@ func Retry[T any](
...
-			log.Info("Context cancelled after attempt %d failed.", attempt)
+			log.Info("Context cancelled after attempt %d failed.", "attempt", attempt)
 			return zero, ctx.Err() // Return context error preferentially
 		default:
 			// Context not cancelled, proceed with error checking
 		}
 
 		if !isRetryable(lastErr) {
-			log.Info("Attempt %d failed with non-retryable error: %v", attempt, lastErr)
+			log.Info("Attempt failed with non-retryable error", "attempt", attempt, "error", lastErr)
```
justinsb, line 164: This would be a good one to see if we could get an LLM to catch :-)  Or a linter!

justinsb (approved):

gollm/factory.go
droot: Ack.

PR #80, opened 2025-03-27
Streaming support

justinsb:
Only implemented for the gemini/vertex client currently.

Also improve concurrency in the document model

### DIFF
--- a/pkg/agent/conversation.go
+++ b/pkg/agent/conversation.go
@@ -151,28 +151,17 @@ func (a *Conversation) RunOneRound(ctx context.Context, query string) error {
 			Payload:   []any{currChatContent},
 		})
 
-		response, err := a.llmChat.Send(ctx, currChatContent...)
+		stream, err := a.llmChat.SendStreaming(ctx, currChatContent...)
 		if err != nil {
 			return err
 		}
 
-		a.Recorder.Write(ctx, &journal.Event{
-			Timestamp: time.Now(),
-			Action:    "llm-response",
-			Payload:   response,
-		})
-
+		// Clear our "response" now that we sent the last response
 		currChatContent = nil
-
-		if len(response.Candidates()) == 0 {
-			return fmt.Errorf("no candidates in LLM response")
-		}
-
-		candidate := response.Candidates()[0]
 
 		if a.EnableToolUseShim {
 			// convert the candidate response into a gollm.ChatResponse
-			candidate, err = candidateToShimCandidate(candidate)
+			stream, err = candidateToShimCandidate(stream)
 			if err != nil {
 				return err
 			}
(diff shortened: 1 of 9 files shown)


pkg/agent/conversation.go
```diff
@@ -182,95 +171,118 @@ func (a *Conversation) RunOneRound(ctx context.Context, query string) error {
...
-					s := toolCall.PrettyPrint()
-					a.doc.AddBlock(ui.NewAgentTextBlock().SetText(fmt.Sprintf("  Running: %s\n", s)).SetColor(ui.ColorGreen))
-					// Ask for confirmation only if SkipPermissions is false AND the tool modifies resources.
-					if !a.SkipPermissions && call.Arguments["modifies_resource"] != "no" {
-						confirmationPrompt := `  Do you want to proceed?
+			s := toolCall.PrettyPrint()
+			a.doc.AddBlock(ui.NewAgentTextBlock().SetText(fmt.Sprintf("  Running: %s\n", s)).SetColor(ui.ColorGreen))
+			// Ask for confirmation only if SkipPermissions is false AND the tool modifies resources.
+			if !a.SkipPermissions && call.Arguments["modifies_resource"] != "no" {
+				confirmationPrompt := `  Do you want to proceed?
```
justinsb, line 229: I think we could ask for a "group" of functions at once.  I don't think we can allow evaluation of some but not all functions anyway, the LLM will complain if the number of function responses does not match the number of function requests.  (Although _maybe_ we can work around that by returning an error like "The user denied permission to run the function")  But in any case, it seems like it might be better UX to show all the functions at once!

pkg/agent/conversation.go
```diff
@@ -151,28 +151,17 @@ func (a *Conversation) RunOneRound(ctx context.Context, query string) error {
 			Payload:   []any{currChatContent},
 		})
 
-		response, err := a.llmChat.Send(ctx, currChatContent...)
+		stream, err := a.llmChat.SendStreaming(ctx, currChatContent...)
```
droot, line 154: Would have been nice if this was configurable. Though not sure how easy it will be to make it configurable.

pkg/agent/conversation.go
droot: I have a change that adds function response for user not approving calling the function so that will take care of this problem. I also tweaked the flow to not exit the repl if user didn't approve.

Showing all functions or even running all them in parallel... we can definitely add those, but at the moment I have seen only 5% cases where more than one function is invoked, so probably not a big deal.

Also, wrt to streaming and calling, in my experience of running gemini-2.5-pro, agent spend more time on executing the function than waiting for llm response :)

droot (approved):

pkg/agent/conversation.go
justinsb: > I have a change that adds function response for user not approving calling the function so that will take care of this problem. I also tweaked the flow to not exit the repl if user didn't approve.

Cool!  Maybe we have an approve all button alongside a function by function approve?

> Showing all functions or even running all them in parallel... we can definitely add those, but at the moment I have seen only 5% cases where more than one function is invoked at the same time, so probably not a big deal.

I've seen it when I ask it to do something e.g. for each pod, like vuln scan all the images my pods are running.

> Also curious, for tool calling, how will streaming response look like ? assuming sequential tool calls, each streaming chunk probably returns single tool call, right ?

I don't think we have any guarantees actually.  I _think_ though that the streaming call is the same as the non-streaming call, you just get your results incrementally.  Same stopping criteria, just that you get results every half second instead of having to wait.

> Also, wrt to streaming and calling, in my experience of running gemini-2.5-pro, agent spend more time on executing the function than waiting for llm response :)

Yeah, gemini is definitely setting the bar really high here in terms of speed :-)

pkg/agent/conversation.go
justinsb: So we already "backfill" support for streaming on most of the providers, so we could easily do that for gemini also.  I think we want to write code that supports streaming in anticipation of slower models, but we could back it with non-streaming requests if they are cheaper / better etc.

PR #82, opened 2025-03-27
Use gemini-2.5-pro as default gemini model

droot:
Two changes:

 - use gemini-2.5-pro-experimental (Google's strongest AI model yet) as the default
 - Use tool-calling by default because the default model supports tool use natively.


janetkuo (approved):
/lgtm

PR #83, opened 2025-03-29
Fixed misc bugs in mcp server code

droot:
- MCP server code was crashing because `trivy` tool's schema is different from the other tools. The change fixes that
- Added a sample configurations doc


mikebz:
small nit - link out the mcp doc from the main readme if it hasn't been

droot:
@justinsb merging this because `MCP` server mode panics today.

PR #89, opened 2025-04-09
gollm: implement registry pattern for parsing providers

justinsb:


janetkuo (approved):

droot (approved):

justinsb:
> My mental model is: model is a request parameter for a llm request, so it is probably better to keep it separate from the URI for the llm provider. URI for llm provider's sole job is to specify the coordinates for endpoint for llm provider.

So the analogy I would give is a connection URL.  Mysql has [this](https://dev.mysql.com/doc/connector-j/en/connector-j-reference-jdbc-url-format.html)  and postgres has [this](https://www.postgresql.org/docs/current/libpq-connect.html#LIBPQ-CONNSTRING-URIS).  The idea is that you should be able to pass in all the parameters in the URI.

The problem with that analogy is that tables are _not_ in the URI.  So are models tables or are they databases?

I'm going to merge and we can see how we feel, this isn't making any real decisions here yet (and we aren't committing ourselves anyway, IMO, because we are not yet making 1.0 type guarantees.)

PR #93, opened 2025-04-15
Toggled ask-for-configuration defaults to confirm by default

droot:
Highlights:

1. `kubectl-ai` will now ask for permission to run `kubectl` commands that modifies resources in user's cluster. Users can override this behavior by using `--skip-permissions` command line flag.
2. Tweaked the confirmation flow to present users will the choice to 1) approve 2) approve and remember the choice for rest of session 3) decline. The option to approve and remember the choice makes the flow less annoying when user is comfortable with the modifying resources and trusts the tool/model more.

The run below shows the confirmation flow:

```sh
$ ./kubectl-ai "double the capacity of nginx application in hello namespace"

  Okay, I can help with that. First, I need to find the deployment for the nginx
  application in the 'hello' namespace and check its current size.

  Running: kubectl get deployment -n hello -o jsonpath="{range .items[?(@.metadata.name==\"nginx\" || contains(@.metadata.name, \"nginx\"))]}{.metadata.name}{'\t'}{.spec.replicas}{'\n'}{end}"

  Okay, I see the previous attempt failed. Let's try finding the specific
  deployment named 'nginx' first and get its current replica count.

  Running: kubectl get deployment nginx -n hello -o jsonpath='{.spec.replicas}'

  Okay, the current deployment 'nginx' in the 'hello' namespace has 1 replica.
  Doubling it means scaling it to 2 replicas.

  Let's scale the deployment now.

  Running: kubectl scale deployment nginx --replicas=2 -n hello
  Do you want to proceed?
  1) Yes
  2) Yes, and don't ask me again
  3) No
  Enter your choice (number): 1

  Done! 🎉 I have scaled the  nginx  deployment in the  hello  namespace to 2
  replicas, effectively doubling its capacity.

```

### DIFF
--- a/pkg/agent/conversation.go
+++ b/pkg/agent/conversation.go
@@ -46,8 +46,8 @@ type Conversation struct {
 
 	MaxIterations int
 
-	Kubeconfig          string
-	AsksForConfirmation bool
+	Kubeconfig      string
+	SkipPermissions bool
 
 	Tools tools.Tools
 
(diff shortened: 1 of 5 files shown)


pkg/agent/conversation.go
```diff
@@ -205,11 +205,26 @@ func (a *Conversation) RunOneRound(ctx context.Context, query string) error {
...
 					s := toolCall.PrettyPrint()
 					a.UI.RenderOutput(ctx, fmt.Sprintf("  Running: %s\n", s), ui.Foreground(ui.ColorGreen))
-					if a.AsksForConfirmation && call.Arguments["modifies_resource"] == "no" {
-						confirm := a.UI.AskForConfirmation(ctx, "  Are you sure you want to run this command (Y/n)? ")
-						if !confirm {
-							a.UI.RenderOutput(ctx, "Sure.\n", ui.RenderMarkdown())
+					// Ask for confirmation only if SkipPermissions is false AND the tool modifies resources.
+					if !a.SkipPermissions && call.Arguments["modifies_resource"] != "no" {
+						confirmationPrompt := `  Do you want to proceed?
+  1) Yes
```
justinsb, line 211: Nit: y / Y / n might be more intuitive, but ... just a nit

justinsb (approved):

pkg/agent/conversation.go
droot: I hear you! I think UX wise, using a terminal UI library that supports mouse etc. will completely eliminate the need for typing and is far superior. For now, keeping this as-is, with the hope that we use TUI.

PR #95, opened 2025-04-16
Reintroduce markdown with streaming support

justinsb:
- **Minimal streaming support**
  Also improve concurrency in the document model
  

- **Reintroduce markdown support with streaming**


justinsb:
So we can suppress streaming inside markdown blocks like this.  I think we are no worse off (other than the complexity).  And then we can have a tview UI or similar that can support markdown with streaming.  Technically all we need to do here is record the cursor position and rewind to print the markdown more incrementally, but doing that feels like we're basically implementing tview.

droot (approved):

PR #96, opened 2025-04-16
gollm: default project and location for vertexai

justinsb:
These are both required, so default the project from gcloud.
We use a hard-coded default of us-central1 for location.

This restores the previous behaviour.

### DIFF
--- a/gollm/gemini.go
+++ b/gollm/gemini.go
@@ -15,6 +15,7 @@
 package gollm
 
 import (
+	"bytes"
 	"context"
 	"encoding/json"
 	"errors"
@@ -24,6 +25,7 @@ import (
 	"net/http"
 	"net/url"
 	"os"
+	"os/exec"
 	"strings"
 
 	"google.golang.org/genai"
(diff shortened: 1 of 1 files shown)


gollm/gemini.go
```diff
@@ -44,12 +46,51 @@ func geminiFactory(ctx context.Context, u *url.URL) (Client, error) {
...
+	// Location is also required
+	if opt.Location == "" {
+		opt.Location = "us-central1"
+		log.Info("defaulted location for vertex client", "location", opt.Location)
+	}
 	return NewGeminiClient(ctx, opt)
 }
 
+// findDefaultGCPProject gets the default GCP project ID from gcloud
+func findDefaultGCPProject(ctx context.Context) (string, error) {
```
droot, line 72: This is more of an informational comment:

I have not been using the package or agent with vertex anywhere yet, but if for reason, we don't want to take on dependency on gcloud, we should probably make this friendly with the environment variables supported by the upstream genai package https://github.com/googleapis/go-genai/blob/6ac4afc0168762dc3b7a4d940fc463cc1854f366/client.go#L148

Realized it also supports `GOOGLE_API_KEY` for the gemini backend.

How or where do you use the vertex provider ?

justinsb: So I use the vertex provider because it uses GCP auth.  So if I'm authenticated to a GCP cluster, in theory I don't need to do anything else to use vertex (other than activate the service).

You are right that we should honor the env vars over fallback to gcloud, but I think the use case for things that talk to GKE is that we can give single-sign-on.

PR #98, opened 2025-04-17
Added Quiet mode

droot:
### Quiet mode

This is relevant when you invoke the agent with the query (or task description) on the command line, for ex. `kubectl-ai "bump up the capacity for nginx app"`. This will now drop the user in repl mode by default unless user specified `kubectl-ai --quiet` to disable non-interactive session. This is primarily motivated by my experience of using the tool. I wired the option all the way to `k8s-bench` as well and that helps if you are debugging any eval failure and having access to the repl with all the LLM chat context is useful.

### Bug fixes
 - While using the agent command in a piped syntax or providing the input from stdin, it wasn't compatible with permission and interactive repl flow. Introduced configuration to use tty in those cases.

### Misc

 - Using `gemin-2.5-pro-preview-03-25` by default because it has higher rate limits in free tier by default. Change is in agent as well as k8s-bench.
 - Minor tweak to the approval flow. If user declines running the command, we now record the decline in the chat history and we found LLM handles much more intelligently rest of the flow.
 - We lost the green coloring for function call, so fixed that.

Note: I will update the README as a follow up to include the `quiet` mode.

### DIFF
--- a/k8s-bench/eval.go
+++ b/k8s-bench/eval.go
@@ -310,6 +310,7 @@ func (x *TaskExecution) runAgent(ctx context.Context) error {
 		"--kubeconfig", x.kubeConfig,
 		"--llm-provider", x.llmConfig.ProviderID,
 		fmt.Sprintf("--enable-tool-use-shim=%t", x.llmConfig.EnableToolUseShim),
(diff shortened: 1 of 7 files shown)


justinsb (approved):

justinsb:
LGTM!

PR #99, opened 2025-04-17
gollm: recognize well-known env vars

justinsb:
Recognize the following env vars for vertexai:

* project: GOOGLE_CLOUD_PROJECT
* location: GOOGLE_CLOUD_LOCATION, GOOGLE_CLOUD_REGION

### DIFF
--- a/gollm/gemini.go
+++ b/gollm/gemini.go
@@ -62,9 +62,26 @@ func vertexaiViaGeminiFactory(ctx context.Context, u *url.URL) (Client, error) {
 
 	// Location is also required
 	if opt.Location == "" {
-		opt.Location = "us-central1"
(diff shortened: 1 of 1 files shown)


justinsb:
I'm not sure about GOOGLE_API_KEY, it isn't a GCP API Key, it is a Gemini API Key...  I don't think it will work for GCP, and I don't think it will work for other google services.

droot:
> I'm not sure about GOOGLE_API_KEY, it isn't a GCP API Key, it is a Gemini API Key... I don't think it will work for GCP, and I don't think it will work for other google services.

Yeah, `GOOGLE_API_KEY` will replace our current environment variable `GEMINI_API_KEY` for the gemini backend. It won't work for vertex backend (GCP).

PR #104, opened 2025-04-18
Fix the error building modelserving images

janetkuo:
This fixes this error `CMake could not find CURL` when running `modelserving/dev/tasks/build-images`


justinsb:
Thanks!

/approve
/lgtm

PR #108, opened 2025-04-22
Pass the LLM  provider URL to the Azure OpenAI client

hakman:
This makes the `AZURE_OPENAI_ENDPOINT` optional.

### DIFF
--- a/gollm/azopenai.go
+++ b/gollm/azopenai.go
@@ -32,7 +32,7 @@ func init() {
 }
 
 func azureOpenAIFactory(ctx context.Context, u *url.URL) (Client, error) {
-	return NewAzureOpenAIClient(ctx)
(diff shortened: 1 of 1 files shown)


gollm/azopenai.go
```diff
@@ -32,7 +32,7 @@ func init() {
 }
 
 func azureOpenAIFactory(ctx context.Context, u *url.URL) (Client, error) {
```
hakman, line 34: Not quite sure why factory needs to receive a pointer to the URL. I would expect a copy to be enough for the clients.
https://github.com/GoogleCloudPlatform/kubectl-ai/blob/b5f05da496cde0cb1c159e1e31d770c5b7f135f2/gollm/factory.go#L39

droot (approved):

PR #110, opened 2025-04-23
k8s-bench: Fix wrong `indirect` k8s.io/klog/v2 in the go.mod

zchee:
The k8s.io/klog/v2 package is not `indirect`. Also cleanup.


justinsb:
Thank you @zchee!  I guess this means we are missing a test (in GitHub actions?) that when we run `go mod tidy` that nothing changes.

/approve
/lgtm

PR #127, opened 2025-04-30
fix: don't dictate the name for the models

cwrau:
If the user has chosen a model, it's probably because the user wanted to use it...


droot (approved):
The change makes sense. Thank you!

PR #129, opened 2025-05-01
Cli/121

selimacerbas:
onboard cobra and viper **scoped** to kubectl-ai. 

What features are added;

1- Removed "flags" pkg dependency and replaced the cli logic with cobra framework.
2- Introduced viper framework to handle kubectl-ai configuration over ENV variables.
3- Introduced KUBECTL_AI as a prefix for all defined ENV variables, users can now define ENV variables like,
`KUBECTL_AI_MODEL=gemini-2.0-flash kubectl-ai` and fire up the tool. 
4- Introduced helper method called isFlagValueCorrect to check if passed user value for the flag is correct to eliminate runtime errors. (For now this is only scoped to --model flag)
5- collected initialisation logic into init() function.

Further Suggestions;
1- Cleanup runCmd method, separate logic into functions.
2- Some variables are defined as global and some are under Options struct. I would suggest collecting all of CLI related variables under CLIOptions Struct removing globally defined variables. This would reduce complexity. 

Here are couple screenshots, 
Screenshot 2025-05-01 at 17 59 07
Screenshot 2025-05-01 at 18 51 57


main.go
```diff
@@ -45,48 +47,39 @@ var (
...
+	removeWorkDir          bool
+	geminiModels           = []string{
+		"gemini-2.5-pro-preview-03-25",
+		"gemini-2.0-flash",
+	}
 
-	if err := run(ctx); err != nil {
-		fmt.Fprintln(os.Stderr, err)
-		os.Exit(1)
+	rootCmd = &cobra.Command{
```
rxinui, line 63: for sake of clarity and to prepare future enhancement, i would define `var rootCmd` on its own before line 50.

main.go
```diff
@@ -153,55 +145,97 @@ func (o *Options) LoadConfigurationFile() error {
...
+	f.BoolVar(&opt.MCPServer, "mcp-server", opt.MCPServer, "run in MCP server mode")
+	f.BoolVar(&opt.EnableToolUseShim, "enable-tool-use-shim", opt.EnableToolUseShim, "enable tool use shim")
+	f.BoolVar(&opt.Quiet, "quiet", opt.Quiet, "run in non-interactive mode, requires a query to be provided as a positional argument")
+
+	// viper binds and env var prefixes
+	err := viper.BindPFlags(f)
+	if err != nil {
+		print(fmt.Errorf("failed to bind viper flags"))
 	}
+	viper.SetEnvPrefix("KUBECTL_AI")
```
rxinui, line 172: I would declare a const for `KUBECTL_AI` at the beginning

main.go
```diff
@@ -214,43 +248,37 @@ func run(ctx context.Context) error {
...
+	switch {
 	// Handle positional arguments and stdin
-	if len(args) > 1 {
-		return fmt.Errorf("only one positional argument (query) is allowed")
-	} else if stdinHasData {
-		// Read from stdin
-		scanner := bufio.NewScanner(os.Stdin)
-		var queryBuilder strings.Builder
-
+	case len(args) > 1:
```
rxinui, line 258: because of cobra introduction, some logic SHOULD be refactored.

Positional args https://github.com/spf13/cobra/blob/main/site/content/user_guide.md#positional-and-custom-arguments should be treated that way instead.

main.go
```diff
@@ -440,3 +468,26 @@ func (writer klogWriter) Write(data []byte) (n int, err error) {
 	klog.Warning(message)
 	return len(data), nil
 }
+
+func getDefaultKubeConfigPath() (string, error) {
```
rxinui, line 472: This is not idiomatic. Best way would be to make use of https://pkg.go.dev/k8s.io/client-go/tools/clientcmd#LoadFromFile 

Some constants are available as well to control the flow.

main.go
```diff
@@ -153,55 +145,97 @@ func (o *Options) LoadConfigurationFile() error {
...
+
+	if err := rootCmd.ExecuteContext(ctx); err != nil {
+		fmt.Fprintln(os.Stderr, err)
+		os.Exit(1)
+	}
+}
+
+func runCmd(cmd *cobra.Command, args []string) error {
+	// Retrieving the context from cmd which passed in init()
+	ctx := cmd.Context()
```
rxinui, line 198: IMO, these vipers instructions should be under a func `configureViper` which is called by `runCmd` to abstract the CLI setting to the CLI logic.

From line 198 til 238

main.go
```diff
@@ -45,48 +47,39 @@ var (
...
-		os.Exit(0)
-	}()
+var (
+	// CLI related
+	opt                    = Options{}
+	maxIterations          int
+	kubeconfigPath         string
+	promptTemplateFilePath string
+	tracePath              string
+	removeWorkDir          bool
```
droot, line 57: I would suggest moving all the above to `Options{}` as well.

main.go
```diff
@@ -45,48 +47,39 @@ var (
...
-
-	go func() {
-		sig := <-sigCh
-		fmt.Fprintf(os.Stderr, "Received signal, shutting down... %s\n", sig)
-		klog.Flush()
-		os.Exit(0)
-	}()
+var (
+	// CLI related
+	opt                    = Options{}
```
rxinui, line 52: best to just declare it to remain consistent `opt Options`

main.go
```diff
@@ -153,55 +145,97 @@ func (o *Options) LoadConfigurationFile() error {
...
+
+	// viper binds and env var prefixes
+	err := viper.BindPFlags(f)
+	if err != nil {
+		print(fmt.Errorf("failed to bind viper flags"))
 	}
+	viper.SetEnvPrefix("KUBECTL_AI")
+	viper.SetEnvKeyReplacer(strings.NewReplacer("-", "_"))
+	viper.AutomaticEnv()
+}
```
droot, line 175: I see that `klog.InitFlags(nil)` is no longer here ? (that is useful for controlling the logging configuration).

main.go
```diff
@@ -45,48 +47,39 @@ var (
...
-	}()
+var (
+	// CLI related
+	opt                    = Options{}
+	maxIterations          int
+	kubeconfigPath         string
+	promptTemplateFilePath string
+	tracePath              string
+	removeWorkDir          bool
+	geminiModels           = []string{
```
rxinui, line 58: same, would just declare the type and set default values elsewhere, surely the init() at that point.

rxinui (changes requested):
As @droot mentioned, I'm reviewing your code. I left some enhancement comments, but overall it looks ok.

main.go
```diff
@@ -153,55 +145,97 @@ func (o *Options) LoadConfigurationFile() error {
...
+	opt.ProviderID = viper.GetString("llm-provider")
+	opt.ModelID = viper.GetString("model")
+	opt.SkipPermissions = viper.GetBool("skip-permissions")
+	opt.MCPServer = viper.GetBool("mcp-server")
+	opt.EnableToolUseShim = viper.GetBool("enable-tool-use-shim")
+	opt.Quiet = viper.GetBool("quiet")
 
-	flag.Parse()
+	// Validation of flag values passed by user, use this helper method
+	_, err := isFlagValueCorrect(opt.ModelID, geminiModels)
```
droot, line 221: Let's avoid validating models because it is not possible to do it correctly without fetching the supported models for a given llm-provider and some llm-providers don't even allow listing models.

Note that we support gemini, openai, vertexai, az-openai, ollama, llaama.cpp as llm-provider today and each support different models.

main.go
```diff
@@ -153,55 +145,97 @@ func (o *Options) LoadConfigurationFile() error {
 	return nil
 }
 
-func run(ctx context.Context) error {
-	// Command line flags
-	var opt Options
+func init() {
```
droot, line 148: would be good to avoid the flag parsing logic out of `init()` block. Let's have this closer to the defaults initialization in the main().

droot (commented):
Thanks for starting on this. I have some suggestions.

main.go
rxinui: I agree, i'm not a fan of using `init` as it is discouraged yet it is the most common pattern for cli. My recommendation is to separate it as functions and call it in main

selimacerbas:
Thanks for the reviews and docu references, I will than move the global variables into Options Struct and remove init() functions. Is there a specific reason that we avoid using `init()` function? I am assuming only to have one entry in the code? Just out of curiosity.

rxinui:
@selimacerbas sure, as references:
- Quick ref: https://100go.co/#misusing-init-functions-3
- Deeper ref (which is how i get to know about it): [search for "Side Effect Set-up"](https://medium.com/capital-one-tech/doing-well-by-doing-bad-writing-bad-code-with-go-part-1-2dbb96ce079a )

In my coding style, I do avoid global var and init function to not have side-effect (i love functional programming style for that reason) so it's just my own rule. 
In case of CLI, we can just replace this `init` function by ad-hoc functions called within `main` which returns the configured command. That is my preference but spf13 himself uses `init` so *who am I to argue...* 🙇

selimacerbas:
@rxinui Very interesting!!! Thanks for the references! Indeed i wanted to return error in init() function and did not work... tbh i did not look from limitations perspective but yeah why not other workarounds🙂 and ad-hoc functions would be a good idea.

main.go
droot: Using `client-go` is very tempting but keeping up with the k8s dependencies is not fun, and especially in this case of just default kubeconfig is not worth it. So I would sugges to keep it as-is.

droot:
> @selimacerbas sure, as references:
> 
> * Quick ref: https://100go.co/#misusing-init-functions-3
> * Deeper ref (which is how i get to know about it): [search for "Side Effect Set-up"](https://medium.com/capital-one-tech/doing-well-by-doing-bad-writing-bad-code-with-go-part-1-2dbb96ce079a)
> 
> In my coding style, I do avoid global var and init function to not have side-effect (i love functional programming style for that reason) so it's just my own rule. In case of CLI, we can just replace this `init` function by ad-hoc functions called within `main` which returns the configured command. That is my preference but spf13 himself uses `init` so _who am I to argue..._ 🙇

+100

main.go
selimacerbas: done

main.go
selimacerbas: I have moved he logic to resolveKubeConfigPath function by beautifying it.

main.go
selimacerbas: I  have separated Flag bindings and Env setup logic of viper. Also created separate functions to populate Options values.

main.go
selimacerbas: moved all of the setting under Options{} struct.

main.go
selimacerbas: removed it

main.go
selimacerbas: removed init() from logic.

selimacerbas:
Here are my updates,

Fixes;
1- Old CLI was not able to parse positional arg and flags at the same time. Here is screenshot before after;
Before;
Screenshot 2025-05-03 at 01 0

After;
<img width=

2-Old CLI was not able to parse query  correctly  if positional arg and stdin is given at the same time.  Here are some screenshots; (Check resolveQueryInput function to see how it  is handled)
Before;
Screenshot 2025-05-03 at 02 01 12

After;
Screenshot 2025-05-03 at 02 01 34

Enhancements;
1- Separated Vipers ENV setup and CLI Flag bindings into two functions.
2- Removed Globally defined values and moved them under Options struct.
3- Created populateOptionsFromViper function to populate Option from cli flags and ENV vars.
4- Brought mistakenly deleted logging initFlag function.
5- Beautified and kubeconfig resolve logic
6- Created resolveQueryInput method to handle all pos args and stdin values. Users can now create create queries like, 
Screenshot 2025-05-03 at 01 29 32
Screenshot 2025-05-03 at 01 32 10
7- Refactored pos arg handling logic as it is defined in cobras docu. @rxinui thanks for the link!

Thanks in advance for the review!

main.go
```diff
@@ -153,105 +146,107 @@ func (o *Options) LoadConfigurationFile() error {
...
 	}
 
-	// Check for positional arguments (after all flags are parsed)
-	args := flag.Args()
-	var queryFromCmd string
+	// populate Options either from CLI flags or ENV vars
+	populateOptionsFromViper(&opt)
+
+	// add commandline flags for logging
+	klog.InitFlags(nil)
```
droot, line 223: This needs to be called before flag parsing is called, so move it before `runCmd` is invoked. I think thats the reason you are seeing logs output by default on stdout/stderr.

main.go
```diff
@@ -263,8 +258,9 @@ func run(ctx context.Context) error {
 	defer llmClient.Close()
 
 	var recorder journal.Recorder
-	if *tracePath != "" {
-		fileRecorder, err := journal.NewFileRecorder(*tracePath)
+	// CHECK: By default it is set trace.log. It will never be empty string.
```
droot, line 261: didn't understand the comment.

droot (changes requested):
Looking very close. couple of minor comments. Can you pl. rebase with the upstream and update the PR, I would like to try out a quick run.

droot:
Re: default trace path
we can default to `filepath.Join(os.TempDir(), "kubectl-ai-trace.txt")`

Thanks for the detailed screenshot. Being able to parse all types of command line args was the intention with this change and seeing it working is great. Thanks

klog output on the default stdout/stdin is an issue, I have commented on the PR why that might be happening, try that out and let me know it goes. That is a blocker.

Thanks.

main.go
```diff
@@ -45,61 +46,53 @@ var (
...
+	MCPServer              bool   `json:"mcpServer,omitempty"`
+	MaxIterations          int    `json:"maxIterations,omitempty"`
+	KubeConfigPath         string `json:"kubeConfigPath,omitempty"`
+	PromptTemplateFilePath string `json:"promptTemplateFilePath,omitempty"`
+	TracePath              string `json:"tracePath,omitempty"`
+	RemoveWorkDir          bool   `json:"removeWorkDir,omitempty"`
 }
 
-func (o *Options) InitDefaults() {
+func (o *Options) OptionsInitial() {
```
droot, line 80: Can we retain the name `InitDefaults` ?

main.go
selimacerbas: moved it to main() function

selimacerbas:
I rebased it and also implemented the last requests, happy testing!

droot (commented):
Awesome work. I just tested it, and working exactly as expected. Some minor comments and cleanup, then we are good to go.

Thank you for awesome work.

main.go
```diff
@@ -153,105 +147,122 @@ func (o *Options) LoadConfigurationFile() error {
...
+	// Config for logs:
+	// — file-only:
+	// flag.Set("logtostderr", "false")
+	// flag.Set("log-file", "/tmp/kubectl-ai.log")
+	//
+	// — or stderr-only: Which is default of klog.
+	// flag.Set("logtostderr", "true")
+
+	flag.Set("logtostderr", "false")
+	flag.Set("log_file", "/tmp/kubectl-ai.log")
```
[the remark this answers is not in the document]
selimacerbas: It is a good point I did not consider it... thanks for the explanation

main.go
```diff
@@ -153,105 +147,122 @@ func (o *Options) LoadConfigurationFile() error {
...
+		os.Exit(1)
+	}
+}
 
-	// add commandline flags for logging
-	klog.InitFlags(nil)
+func bindCLIFlagsToViper(opt Options) error {
+	f := rootCmd.Flags()
+	f.IntVar(&opt.MaxIterations, "max-iterations", opt.MaxIterations, "maximum number of iterations agent will try before giving up")
+	f.StringVar(&opt.KubeConfigPath, "kubeconfig-path", opt.KubeConfigPath, "path to kubeconfig file")
```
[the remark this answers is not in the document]
selimacerbas: sure it is renamed.

selimacerbas:
Happy to hear that everything is working fine! Thank you for the reviews and support @rxinui and @droot !

rxinui:
I'm happy I got to help the project and at the same time, influence your go skill towards the right direction @selimacerbas 😉

droot:
> I'm happy I got to help the project and at the same time, influence your go skill towards the right direction @selimacerbas 😉

This right there is what OSS is about! Appreciate you both helping with the project. Thank you and cheers! 🚢

PR #132, opened 2025-05-02
use os.TempDir()

mattn:
Sorry, I find one more /tmp


justinsb:
Thanks @mattn.  We should probably find a linter that can detect path.Join and usage of e.g. "/tmp".  I suspect that's tricky to do "classically", but it might be fun to see whether LLMs can flag these things.

But that's a tangent :-) ...

/approve
/lgtm

PR #138, opened 2025-05-05
fixed the use of emojis in systemprompt_template

mschneider82:
there was a typo in emojis in the systemprompt.


droot (approved):
Thank you.

PR #141, opened 2025-05-05
fix(tmp): correct overwritten tmp folder mentioned in (#132)

selimacerbas:
I noticed that the change of @mattn in this commit https://github.com/GoogleCloudPlatform/kubectl-ai/commit/5fd397b2cf153798f30209fff191258c48289433 did not take effect since there was another change on top of it. I raised this PR to fix it.


droot (approved):
Good catch! Thank you.

PR #148, opened 2025-05-07
Omit dollar signs

sinwoobang:
$ is not part of the command, but a prompt indicator. It is better to avoid including the shell prompt symbol ($) so users can copy commands easily.

Tools like markdownlint support style rules such as [MD014 – Dollar signs used before commands without showing output](https://github.com/DavidAnson/markdownlint/blob/main/doc/Rules.md#md014), encouraging separation of prompt and command.


droot (approved):
Thanks. Makes sense.

PR #150, opened 2025-05-07
feat(install): add quick and manual install options in README and script

tuannvm:
## Summary
This update introduces a new `install.sh` script into the repository, enabling streamlined installation of the `kubectl-ai` CLI tool. The script automatically detects the user's OS and architecture, retrieves the latest release from GitHub, and installs the binary into `/usr/local/bin`. This simplifies the setup process, ensuring users can easily deploy the CLI tool with minimal manual steps.

## Affected Modules
- New `install.sh` script added to repository root.

## Key Details
- **OS Detection:** Supports Linux and macOS (Darwin).
- **Architecture Detection:** Supports amd64/x86_64 and arm64/aarch64.
- **Latest Release Fetch:** Uses the GitHub API to automatically identify the most recent release.
- **Installation Method:** Downloads the release tarball, extracts the binary, and installs it to `/usr/local/bin`, with required `sudo` privileges.
- **User Feedback:** Provides clear messaging on download progress, installation status, and success confirmation.

## Potential Impacts
- Provides a quick, reliable method for users to install `kubectl-ai`.
- Facilitates consistency across environments, reducing setup errors.
- Minimal risk; installation relies on standard directory paths and permissions.

### DIFF
--- a/README.md
+++ b/README.md
@@ -16,7 +16,7 @@ First, ensure that kubectl is installed and configured.
 curl -sSL https://raw.githubusercontent.com/GoogleCloudPlatform/kubectl-ai/main/install.sh | bash
 ```
 
-#### Manual Installation
+#### Manual Installation (Linux, MacOS and Windows)
 
 1. Download the latest release from the [releases page](https://github.com/GoogleCloudPlatform/kubectl-ai/releases/latest) for your target machine.
 
--- a/README.md
+++ b/README.md
@@ -10,6 +10,14 @@ First, ensure that kubectl is installed and configured.
 
 ### Installation
 
+#### Quick Install
+
+```shell
+curl -sSL https://raw.githubusercontent.com/GoogleCloudPlatform/kubectl-ai/main/install.sh | bash
+```
+
+#### Manual Installation
+
 1. Download the latest release from the [releases page](https://github.com/GoogleCloudPlatform/kubectl-ai/releases/latest) for your target machine.
 
 2. Untar the release, make the binary executable and move it to a directory in your $PATH (as shown below).
--- a/README.md
+++ b/README.md
@@ -10,7 +10,7 @@ First, ensure that kubectl is installed and configured.
 
 ### Installation
 
-#### Quick Install
+#### Quick Install (Linux & MacOS only)
 
 ```shell
 curl -sSL https://raw.githubusercontent.com/GoogleCloudPlatform/kubectl-ai/main/install.sh | bash
--- a/install.sh
+++ b/install.sh
@@ -1,5 +1,5 @@
 #!/usr/bin/env bash
-set -e
+set -euo pipefail
 
 REPO="GoogleCloudPlatform/kubectl-ai"
 BINARY="kubectl-ai"
@@ -19,7 +19,7 @@ esac
 # Detect ARCH
 ARCH="$(uname -m)"
 case "$ARCH" in
-  x86_64|amd64) ARCH="amd64" ;;
+  x86_64|amd64) ARCH="x86_64" ;;
   arm64|aarch64) ARCH="arm64" ;;
   *)
     echo "If you are on an unsupported architecture, please follow the manual installation instructions at:"
@@ -28,8 +28,15 @@ case "$ARCH" in
     ;;
 esac
 
-# Get latest version tag from GitHub API (portable, no grep -P)
-LATEST_TAG=$(curl -s "https://api.github.com/repos/$REPO/releases/latest" | sed -n 's/.*"tag_name": *"\([^"]*\)".*/\1/p')
+# Get latest version tag from GitHub API, Use GITHUB_TOKEN if available to avoid potential rate limit
+if [ -n "${GITHUB_TOKEN:-}" ]; then
+  auth_hdr="Authorization: token $GITHUB_TOKEN"
+else
+  auth_hdr=""
+fi
+LATEST_TAG=$(curl -s -H "$auth_hdr" \
+  "https://api.github.com/repos/$REPO/releases/latest" \
+  | sed -n 's/.*"tag_name": *"\([^"]*\)".*/\1/p')
 if [ -z "$LATEST_TAG" ]; then
   echo "Failed to fetch latest release tag."
   exit 1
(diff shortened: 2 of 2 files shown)


tuannvm:
Local test

```
λ kubectl-ai % curl -sSL https://raw.githubusercontent.com/tuannvm/kubectl-ai/docs/one-liner-installation/install.sh | bash

Downloading https://github.com/GoogleCloudPlatform/kubectl-ai/releases/download/v0.0.7/kubectl-ai_darwin_arm64.tar.gz ...
  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed
  0     0    0     0    0     0      0      0 --:--:-- --:--:-- --:--:--     0
100  9.7M  100  9.7M    0     0  16.4M      0 --:--:-- --:--:-- --:--:--  103M
Installing kubectl-ai to /usr/local/bin (may require sudo)...
✅ kubectl-ai installed successfully! Run 'kubectl-ai --help' to get started.
```

install.sh
```diff
@@ -0,0 +1,50 @@
...
+
+REPO="GoogleCloudPlatform/kubectl-ai"
+BINARY="kubectl-ai"
+
+# Detect OS
+OS="$(uname | tr '[:upper:]' '[:lower:]')"
+case "$OS" in
+  linux)   OS="linux" ;;
+  darwin)  OS="darwin" ;;
+  *) echo "Unsupported OS: $OS"; exit 1 ;;
```
janetkuo, line 12: `kubectl-ai` supports windows as well

install.sh
```diff
@@ -0,0 +1,50 @@
...
+tar -xzf "$TARBALL"
+
+# Move binary to /usr/local/bin (may require sudo)
+echo "Installing $BINARY to /usr/local/bin (may require sudo)..."
+sudo mv "$BINARY" /usr/local/bin/
+
+# Ensure the binary is executable
+sudo chmod +x /usr/local/bin/$BINARY
+
+# Clean up
```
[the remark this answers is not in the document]
tuannvm: that step is handled by MacOS as far as I know.

install.sh
tuannvm: Yea I don't have Windows machine to test (unless @mattn is interested 😄) I would change the wording a bit to clarify this method is for Linux & MacOS

tuannvm:
@droot please have another look. Thanks!

emlaker:
Really nice work! Auto-detecting OS and architecture definitely smooths out the install process. Just curious as an observer: what happens if someone runs this on an unsupported OS with something more niche? Does the script handle that with a clear message?

tuannvm:
@emlaker https://github.com/GoogleCloudPlatform/kubectl-ai/pull/150/files#diff-043df5bdbf6639d7a77e1d44c5226fd7371e5259a1e4df3a0dd5d64c30dca44fR13-R15

README.md
```diff
@@ -10,6 +10,14 @@ First, ensure that kubectl is installed and configured.
 
 ### Installation
 
+#### Quick Install (Linux & MacOS only)
+
+```shell
+curl -sSL https://raw.githubusercontent.com/GoogleCloudPlatform/kubectl-ai/main/install.sh | bash
+```
+
+#### Manual Installation
```
droot, line 19: can we add `(Linux, MacOS and Windows)` here. I know it is verbose, but do want users to think that windows is not supported.

install.sh
```diff
@@ -0,0 +1,58 @@
...
+    echo "If you are on Windows or another unsupported OS, please follow the manual installation instructions at:"
+    echo "https://github.com/GoogleCloudPlatform/kubectl-ai#manual-installation"
+    exit 1
+    ;;
+esac
+
+# Detect ARCH
+ARCH="$(uname -m)"
+case "$ARCH" in
+  x86_64|amd64) ARCH="amd64" ;;
```
droot, line 22: releases URL expects `ARCH` to be `x86_64` instead of `amd64`.

install.sh
```diff
@@ -0,0 +1,58 @@
...
+echo "Downloading $URL ..."
+curl -LO "$URL"
+tar -xzf "$TARBALL"
+
+# Move binary to /usr/local/bin (may require sudo)
+echo "Installing $BINARY to /usr/local/bin (may require sudo)..."
+sudo mv "$BINARY" /usr/local/bin/
+
+# Ensure the binary is executable
+sudo chmod +x /usr/local/bin/$BINARY
```
droot, line 52: nit: we can make the binary executable right after untaring it to avoid using `sudo` just for that. Also, I think we don't need to do this because untaring results into an executable binary,

install.sh
```diff
@@ -0,0 +1,58 @@
...
+
+# Get latest version tag from GitHub API (portable, no grep -P)
+LATEST_TAG=$(curl -s "https://api.github.com/repos/$REPO/releases/latest" | sed -n 's/.*"tag_name": *"\([^"]*\)".*/\1/p')
+if [ -z "$LATEST_TAG" ]; then
+  echo "Failed to fetch latest release tag."
+  exit 1
+fi
+
+# Compose download URL
+TARBALL="kubectl-ai_${OS}_${ARCH}.tar.gz"
```
droot, line 39: One thing I noticed that the OS name in the tar file is capitalized `Darwin`, `Linux`, `Windows`, but here we will be using OS name all lower-cased. I wonder if it will work with curl on all OSes. Lets try to match them exact to avoid any edge cases.

I think this script is super important because failures here will frustrate the users and some of them may never bother to install and just abort it. So better to be safe than sorry.

install.sh
```diff
@@ -0,0 +1,58 @@
...
+  x86_64|amd64) ARCH="amd64" ;;
+  arm64|aarch64) ARCH="arm64" ;;
+  *)
+    echo "If you are on an unsupported architecture, please follow the manual installation instructions at:"
+    echo "https://github.com/GoogleCloudPlatform/kubectl-ai#manual-installation"
+    exit 1
+    ;;
+esac
+
+# Get latest version tag from GitHub API (portable, no grep -P)
```
droot, line 31: I think we should check if `curl` is available on the system to make the script robust.

droot (changes requested):
Thanks for the update, I have a few more comments.

install.sh
tuannvm: I agree. While URLs are case-insensitive, it would be a good idea to make them 100% matched just in case.

droot (approved):

droot:
Thanks @tuannvm, this is looking great. I will merge this tomorrow.

janetkuo (approved):
/lgtm

PR #151, opened 2025-05-07
Fixes for Windows

### LINKED ISSUE #139: ~/.kube/config is not used ?
I'm using Oracle Cloud Kubernetes (OKI). OKI can be accessed with oci cli command. 

https://docs.oracle.com/en-us/iaas/Content/ContEng/Tasks/contengdownloadkubeconfigfile.htm

In such a configuration, kubectl-ai apparently does not execute the command. kubectl-ai reply nothing for kubectl commands.

mattn:
fixes #139

### DIFF
--- a/pkg/tools/bash_tool.go
+++ b/pkg/tools/bash_tool.go
@@ -17,9 +17,10 @@ package tools
 import (
 	"bytes"
 	"context"
-	"fmt"
 	"os"
 	"os/exec"
+	"path/filepath"
+	"runtime"
 	"strings"
 
 	"github.com/GoogleCloudPlatform/kubectl-ai/gollm"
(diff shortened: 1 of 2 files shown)


pkg/tools/bash_tool.go
```diff
@@ -36,12 +37,13 @@ const (
...
-		if err != nil {
-			return "", err
+		if len(value) >= 2 && value[0] == '~' && os.IsPathSeparator(value[1]) {
+			if runtime.GOOS == "windows" {
+				value = filepath.Join(os.Getenv("USERPROFILE"), value[2:])
+			} else {
+				value = filepath.Join(os.Getenv("HOME"), value[2:])
+			}
 		}
-		return strings.TrimSpace(string(output)), nil
```
mattn, line 44: It may be better to prioritize the expansion of environment variables over the expansion of `~`. What do you think?

droot (approved):
Thank you such for this fix @mattn .
I was initially little hesitant to create releases for windows because we had no way to test on windows and I am glad that we took the risk. So thank you!

PR #177, opened 2025-05-08
feat(tools): add support for streaming command output in bash_tool.go

### LINKED ISSUE #122: FR: please support streaming of logs
was in a situation where I had a workload I wanted to watch when it was starting up (ie tailing its logs). this workload happend to take a long time to become scheduled/runnable, so i was hoping `kubectl ai` could support a prompt along the lines of the following:

```
could you please wait until the workload-with-warmup deployment is running, then tail its logs
```

it is worth noting the following works just as intended:

```
could you please wait until the workload-with-warmup deployment is running, then cat its logs
```

it prints the logs and even remarks that there is just one line at that point:

```
>>> could you please wait until the workload-with-warmup deployment is running, then cat its logs

  Okay, I will wait for the  workload-with-warmup  deployment to become available 
  and then fetch its logs. This might take a moment.                              

  Running: kubectl wait --for=condition=Available deployment/workload-with-warmup --timeout=120s

  The deployment  workload-with-warmup  is now available! ✅                      
                                                                                  
  Now, I'll fetch the logs for you.                                               

  Running: kubectl logs deployment/workload-with-warmup

  Okay, the deployment  workload-with-warmup  is running and here are the latest  
  logs from its default container ( workload ):                                   
                                                                                  
    Mon Apr 28 18:41:50 UTC 2025                                                  
                                                                                  
  It seems there's only one log line since the deployment became available. Let me
  know if you need anything else! 😊                                              

```

when I asked it to tail the logs, it does this which is remarkable, but doesn't handle the streaming part:

```
$ kubectl ai 

  Hey there, what can I help you with today?                                      



>>> could you please wait until the workload-with-warmup deployment is running, then tail its logs           

  Okay, I can do that. I'll wait for the deployment  workload-with-warmup  to     
  become                                                                          
  available and then start streaming its logs.                                    


  Running: kubectl wait --for=condition=available deployment/workload-with-warmup --timeout=300s

  Running: kubectl logs -f deployment/workload-with-warmup

```
it just never returns.

I get this is prob a challenge to handle the streaming, just filing a FR to capture the useful CUJ.

here is the config I had for my "workload":

```yaml
apiVersion: apps/v1
kind: Deployment
metadata:
  name: workload-with-warmup
spec:
  replicas: 1
  selector:
    matchLabels:
      app: workload-with-warmup
  template:
    metadata:
      labels:
        app: workload-with-warmup
    spec:
      initContainers:
        - name: warmup
          image: ubuntu:latest
          args: ['bash', '-c', 'for i in `seq 1 30`; do echo warming up $i; sleep 1s; done']
      containers:
        - name: workload
          image: ubuntu:latest
          args: ['bash', '-c', 'while true; do sleep 3s && date; done']
```

Vinay-Khanagavi:
Adds support for streaming command output in bash_tool.go for commands like kubectl logs -f and tail -f.

- Detects streaming commands by checking for -f or --follow flags.
- Streams stdout/stderr in real time to the user.
Prints a message:

`Streaming output... Press CTRL-C to stop streaming and return to the prompt.
`

- Handles CTRL-C gracefully, returning to the REPL without exiting the tool.
- Non-streaming commands retain their previous behavior.

This improves the UX for monitoring long-running or initializing workloads.

Closes #122

### DIFF
--- a/pkg/tools/bash_tool.go
+++ b/pkg/tools/bash_tool.go
@@ -15,127 +15,205 @@
 package tools
 
 import (
-	"bytes"
-	"context"
-	"os"
-	"os/exec"
-	"path/filepath"
-	"runtime"
-	"strings"
+    "bufio"
+    "bytes"
+    "context"
+    "fmt"
+    "os"
+    "os/exec"
+    "os/signal"
+    "path/filepath"
+    "runtime"
+    "strings"
 
-	"github.com/GoogleCloudPlatform/kubectl-ai/gollm"
+    "github.com/GoogleCloudPlatform/kubectl-ai/gollm"
 )
 
 func init() {
-	RegisterTool(&BashTool{})
+    RegisterTool(&BashTool{})
 }
 
 const (
-	bashBin = "/bin/bash"
+    bashBin = "/bin/bash"
 )
 
 // expandShellVar expands shell variables and syntax using bash
 func expandShellVar(value string) (string, error) {
-	if strings.Contains(value, "~") {
-		if len(value) >= 2 && value[0] == '~' && os.IsPathSeparator(value[1]) {
-			if runtime.GOOS == "windows" {
-				value = filepath.Join(os.Getenv("USERPROFILE"), value[2:])
-			} else {
-				value = filepath.Join(os.Getenv("HOME"), value[2:])
-			}
-		}
-	}
-	return os.ExpandEnv(value), nil
+    if strings.Contains(value, "~") {
+        if len(value) >= 2 && value[0] == '~' && os.IsPathSeparator(value[1]) {
+            if runtime.GOOS == "windows" {
+                value = filepath.Join(os.Getenv("USERPROFILE"), value[2:])
+            } else {
+                value = filepath.Join(os.Getenv("HOME"), value[2:])
+            }
+        }
+    }
+    return os.ExpandEnv(value), nil
 }
 
 type BashTool struct{}
 
 func (t *BashTool) Name() string {
-	return "bash"
+    return "bash"
 }
 
 func (t *BashTool) Description() string {
-	return "Executes a bash command. Use this tool only when you need to execute a shell command."
+    return "Executes a bash command. Use this tool only when you need to execute a shell command."
 }
 
 func (t *BashTool) FunctionDefinition() *gollm.FunctionDefinition {
-	return &gollm.FunctionDefinition{
-		Name:        t.Name(),
-		Description: t.Description(),
-		Parameters: &gollm.Schema{
-			Type: gollm.TypeObject,
-			Properties: map[string]*gollm.Schema{
-				"command": {
-					Type:        gollm.TypeString,
-					Description: `The bash command to execute.`,
-				},
-				"modifies_resource": {
-					Type: gollm.TypeString,
-					Description: `Whether the command modifies a kubernetes resource.
+    return &gollm.FunctionDefinition{
+        Name:        t.Name(),
+        Description: t.Description(),
+        Parameters: &gollm.Schema{
+            Type: gollm.TypeObject,
+            Properties: map[string]*gollm.Schema{
+                "command": {
+                    Type:        gollm.TypeString,
+                    Description: `The bash command to execute.`,
+                },
+                "modifies_resource": {
+                    Type: gollm.TypeString,
+                    Description: `Whether the command modifies a kubernetes resource.
 Possible values:
 - "yes" if the command modifies a resource
 - "no" if the command does not modify a resource
 - "unknown" if the command's effect on the resource is unknown
 `,
-				},
-			},
-		},
-	}
+                },
+            },
+        },
+    }
 }
 
 func (t *BashTool) Run(ctx context.Context, args map[string]any) (any, error) {
-	kubeconfig := ctx.Value("kubeconfig").(string)
-	workDir := ctx.Value("work_dir").(string)
-	command := args["command"].(string)
+    kubeconfig := ctx.Value("kubeconfig").(string)
+    workDir := ctx.Value("work_dir").(string)
+    command := args["command"].(string)
 
-	if strings.Contains(command, "kubectl edit") {
-		return &ExecResult{Error: "interactive mode not supported for kubectl, please use non-interactive commands"}, nil
-	}
-	if strings.Contains(command, "kubectl port-forward") {
-		return &ExecResult{Error: "port-forwarding is not allowed because assistant is running in an unattended mode, please try some other alternative"}, nil
-	}
+    if strings.Contains(command, "kubectl edit") {
+        return &ExecResult{Error: "interactive mode not supported for kubectl, please use non-interactive commands"}, nil
+    }
+    if strings.Contains(command, "kubectl port-forward") {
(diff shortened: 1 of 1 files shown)


pkg/tools/bash_tool.go
```diff
@@ -122,12 +125,86 @@ type ExecResult struct {
 }
 
 func executeCommand(cmd *exec.Cmd) (*ExecResult, error) {
+	// Detect streaming commands (e.g., kubectl logs -f, tail -f)
+	isStreaming := false
+	for _, arg := range cmd.Args {
+		if arg == "-f" || arg == "--follow" {
```
droot, line 131: I suspect it will treat `kubectl apply -f` a streaming command. Probably better to be very specific here and we can check for presense of ("kubectl logs" and "-f") or ("kubectl tail" and "-f") ...and so on...

pkg/tools/bash_tool.go
```diff
@@ -122,12 +125,86 @@ type ExecResult struct {
 }
 
 func executeCommand(cmd *exec.Cmd) (*ExecResult, error) {
+	// Detect streaming commands (e.g., kubectl logs -f, tail -f)
```
droot, line 128: it will be good to add the original github issue that prompted this change.

pkg/tools/bash_tool.go
```diff
@@ -122,12 +125,86 @@ type ExecResult struct {
...
+			isStreaming = true
+			break
+		}
+	}
+
+	results := &ExecResult{}
+
+	if isStreaming {
+		// Print streaming message
+		fmt.Fprintln(os.Stdout, "\nStreaming output... Press CTRL-C to stop streaming and return to the prompt.")
```
droot, line 141: Avoid writing to the `os.Stdout` directly. Use the `ui` instance to write to the screen. You will have to make `ui` instance available from the caller to all the way here.

pkg/tools/bash_tool.go
```diff
@@ -122,12 +125,86 @@ type ExecResult struct {
...
+		interrupt := make(chan os.Signal, 1)
+		signal.Notify(interrupt, os.Interrupt)
+		defer signal.Stop(interrupt)
+
+		// Read stdout and stderr concurrently
+		stdoutDone := make(chan struct{})
+		stderrDone := make(chan struct{})
+
+		go func() {
+			scanner := bufio.NewScanner(stdoutPipe)
```
droot, line 169: In order to support both `interactive` and `non-interactive` mode, we had to open tty device when input was provided on stdin. I suspect this will also run into similar issues.

droot (changes requested):
Thanks @Vinay-Khanagavi for taking a crack at it,

I have some comments.

pkg/tools/bash_tool.go
Vinay-Khanagavi: For streaming detection, do you prefer I check specifically for commands like ( "kubectl logs -f") or ("kubectl tail -f") (i.e., both the base command and the -f flag), instead of just any command with -f?
Let me know if you have a preferred way to match these patterns in the codebase.

pkg/tools/bash_tool.go
Vinay-Khanagavi: Yup! I’ll add a reference to the original GitHub issue (e.g., Closes #122) in the code comments, 
Let me know if you have a preferred location for this reference (code comment vs. PR body).

pkg/tools/bash_tool.go
Vinay-Khanagavi: Thanks for pointing out the need to use the ui instance instead of writing directly to os.Stdout.

Do you have any suggestions or **examples** from the codebase on the best way to pass the ui instance down to executeCommand (and the goroutines), especially since it’s currently not available at this level?

Would you recommend passing it as a parameter, using context, or is there a preferred pattern in this project?

pkg/tools/bash_tool.go
Vinay-Khanagavi: My plan is to follow the same approach used for stdin—checking if a TTY is needed and opening /dev/tty when appropriate, so streaming works smoothly in both modes.

what do you think!!! Is this the best way to handle TTY in this project, or do you recommend a different pattern?
Would appreciate your input or approval before I proceed!

droot:
@Vinay-Khanagavi really swamped right now, I will get to this PR tomorrow, sorry for the delay.

pkg/tools/bash_tool.go
```diff
@@ -15,127 +15,205 @@
...
+
+        if err := cmd.Start(); err != nil {
+            results.Error = err.Error()
+            return results, err
+        }
+
+        // Channel to catch interrupt
+        interrupt := make(chan os.Signal, 1)
+        signal.Notify(interrupt, os.Interrupt)
+        defer signal.Stop(interrupt)
```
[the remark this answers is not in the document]
Vinay-Khanagavi: Thanks for raising this!
I've updated the code to proactively block not just kubectl get ... -w but also other blocking/streaming commands like kubectl logs ... -f. Now, if the model tries to run a command with -w or -f in unattended mode, it will return a clear error message instead of running the command and blocking the session.

So, you won’t see the infinite streaming or need to manually interrupt anymore.
If you spot any other kubectl commands that can block like this, let me know and I’ll add them to the block list as well!

Let me know if you have any other suggestions or if you want to test further. Thanks for reviewing!

Vinay-Khanagavi:
Thank you for the review.
@droot  — I've documented the tests I ran with kubectl-ai in the following doc:
[Testing Report](https://docs.google.com/document/d/1S5uec5d2RmrVLV677UfwPMrJCaTWUE7PzbDhY7d1wYE/edit?usp=sharing)

Please feel free to review and suggest any additional changes if needed. Thank you!

zvdy:
Another more complex edge case would be `exec -it bash `, maybe the is out of scope.

Also, my idea was to allow the cli to stream the messages "infinitely" until user wants to stop and feed it back to the ai so it can work upon the data, analyse and take action. 

What do you think? @Vinay-Khanagavi @droot

Vinay-Khanagavi:
This could be a game-changer for Kubernetes troubleshooting and self-healing, but will require careful design for safety, cost, and user trust. Would love to hear your perspectives on scope, priorities, and potential pilot features!

@zvdy @droot  — would appreciate your thoughts!

I've been thinking about this as well. My main concerns center on security, cost, and ensuring that users retain full control over any automated actions.

droot:
Sorry @Vinay-Khanagavi , This PR has turned out to be more complex than I thought. Thanks @zvdy for finding the edge cases and input.

This PR is challenging us into another area of agentic design (collab with a human and how to share control of interaction with the environment as well as I/O) -- when agent requires user's attention (blocking command), it needs to pass the control to the end-user for that action.  (think browser based agent, that requires asking user to login etc.)

I would like us to do the following:

1. Audit all kubectl commands to discover the edge cases
2. Agree on the agent's behavior (from the perspective of the end-user's goal, sometime running a blocking command is not expected by the end-user and sometime blocking command is an exception and control needs to be passed to the end-user). Here going through 3-4 concrete CUJs (user journeys) helps. 
3. And then proceed with the final implementation.

Looking at the PR, I think we will have good handle on the technical aspect, but more work is needed on the audit and behavior definition.

PR #186, opened 2025-05-09
fix Observable#AddSubscription on s is shadowed

anyoptional:
The variable s is defined again inside the for-loop, which causes the external variable s to be hidden, which in turn causes the for-loop to be handled incorrectly.


droot (approved):
Thanks for the fix.

PR #192, opened 2025-05-09
Chore: add existence check for build-images script

seunggihong:
### Summary

This PR improves the robustness of the `run-local` script used for building and launching the local `llama-server`. It adds a safety check to ensure that the required `dev/tasks/build-images` script exists and is executable before attempting to run it.

### Change Details

- ✅ Added an existence and executability check:
  ```bash
  [[ -x dev/tasks/build-images ]] || {
    echo "build-images not found"; exit 1;
  }
  ```
  
 •	Prevents obscure failures in environments where the build script may be missing or not yet generated.
•	Makes the development workflow more predictable and user-friendly.

❓Why

Without this check, the script would fail with a less descriptive error if build-images is missing or lacks execution permission. This change provides an early and clear error message, improving developer experience.

No functional behavior of the build itself is changed.

### DIFF
--- a/modelserving/dev/tasks/run-local
+++ b/modelserving/dev/tasks/run-local
@@ -12,6 +12,11 @@ cd "${SRC_DIR}"
 export ARCHITECTURES=cpu # TODO: we could support cuda locally, but it is slow to build..
 
 # Build and export the docker image; so we are consistent in how we build
+[[ -x dev/tasks/build-images ]] || {
+  echo "ERROR: build-images script not found or not executable"
+  exit 1
+}
+
 mkdir -p .build/llamacpp-server-cpu
 BUILDX_ARGS="--output type=local,dest=.build/llamacpp-server-cpu" dev/tasks/build-images
 


mozartttt:
This is a solid and low-friction robustness improvement.

✅ What's Good
The use of [[ -x dev/tasks/build-images ]] || { ... } is concise and effective.
It ensures the script:

Exists (-e implied within -x)

Is executable — which is often missed in cross-platform or freshly cloned environments.

The early failure with a clear message ("build-images not found") eliminates the classic "permission denied" or "file not found" traps later in the flow.

By exiting immediately, you’re failing fast, which is exactly what you want in local dev tooling.

👨‍💻 Developer Experience Impact
This kind of check prevents time wasted debugging opaque script errors — especially valuable when onboarding new contributors or switching environments.

It also reflects defensive scripting best practices — treating local tooling as part of the codebase quality bar.

🔍 Just One Suggestion
If you want to future-proof it slightly more (in case paths move or symlinks come into play), you could add a more descriptive error, like:

bash
Copiar
Editar
echo "ERROR: Required script 'dev/tasks/build-images' is missing or not executable. Make sure the repo is properly cloned and the script has exec permissions."
Otherwise, it’s 🔒 clean, safe, and transparent.

PR #198, opened 2025-05-09
introduce cmd directory for the CLI code

droot:
Changes to make it easier for new contributors to understand the repo:

- Moving the code related to `kubectl-ai` cmd under `cmd/` directory.
- Added brief notes about various directories.

*Note: my current view about the repo is that its not very hard to move things around, so we should NOT try to aim for perfect (optimal) directory structure, instead optimize for evolving the structure over time as the project need evolve over time. *

If people have ideas about better organization, let me know. 

/cc @mikebz


justinsb:
Agree with moving the "secondary binaries" like the MCP server - I think it will help them to get _more_ exposure.

One thing we are losing here is the ability to `go run github.com/GoogleCloudPlatform/kubectl-ai`.  OTOH I don't think `go run` can follow GitHub repo moves (though I'm not sure), and at the end of the day I think it becomes `go run github.com/GoogleCloudPlatform/kubectl-ai/cmd/kubectl-ai`, which is not _that_ much harder

Therefore...
/approve
/lgtm

justinsb (approved):

mikebz:
thank you for doing this.  I think it improves readability of the repo.

Issue #196, opened 2025-05-09
Establish a robust testing framework for scalability and reliability


janetkuo:
With `kubectl-ai` recently achieving significant traction, it's crucial to establish a robust and comprehensive testing framework. This will ensure the project's continued stability, reliability, and scalability as we add new features and attract more contributors. 

Tasks include but not limited to:

-   [ ] Comprehensive unit tests: to ensure individual components, functions, and logic are thoroughly tested in isolation. This includes, but is not limited to, command parsing, prompt generation logic, Kubernetes API interaction helpers, output formatting, and utility functions.
-   [ ] Integration tests: Validate interactions between `kubectl-ai`'s components and its external dependencies where appropriate.
-   [ ] Mock frameworks that simulate LLM responses to test without real LLMs

droot:
Looking at what we have so far, I think the best will be running a subset of our evals (difficulty level easy) on every PR. Evals are nothing but e2e tests.

Running these evals against vertexai gemini models using [keyless github action](https://cloud.google.com/blog/products/identity-security/enabling-keyless-authentication-from-github-actions?e=48754805) would be ideal and a good starting point.

I will try to get this going this week.

tuannvm:
Additionally, integrating Kubernetes Kind into CI pipeline and running the evaluations on top of it is another idea. This approach would eliminate the need to manually run eval on local machines, allowing us to receive immediate results whenever a new eval is added.

tuannvm:
@droot @janetkuo I’m happy to help with certain parts. Just let me know. Thanks!

janetkuo:
@tuannvm using KinD sounds good! There could also be some automation for updating eval results (likely not after every PR submission, but instead have some criteria for when it'll get updated).

droot:
> [@droot](https://github.com/droot) [@janetkuo](https://github.com/janetkuo) I’m happy to help with certain parts. Just let me know. Thanks!

Awesome. I have made some progress on using [keyless github action](https://cloud.google.com/blog/products/identity-security/enabling-keyless-authentication-from-github-actions?e=48754805) in a GCP project using github action. Found a bug as well in doing this.

I will have the draft PR up and then we can take a look at the gaps where you can help.

Also, I am wondering if there are any llm providers (openrouter and such) that offers credits for OSS projects to run evals.

droot:
[...]

Issue #158, opened 2025-05-07
[Feature Request] goroutines for k8s-bench task execution


zvdy:
# Feature 

Implement a more efficient way to execute and run tasks that is not sequential, relying in goroutines when a `TasksDir` is set.
 
1. Improve the process of benchmarking LLMs and their respective models in a quicker manner.
2. Implement this in the future in CI/CD before main merging, adding a few github action secrets and spinning up a mock cluster should do the job as far as I know

Hope this makes sense.

# Bottlenecks 

One thing worth noting is that the _"simple"_ solution which is already implemented  executes tasks sequentially, which will take quite some time depending on the hardware and the amount of tasks the folder contains, also based on #145 , I agree It's missing some common troubleshooting/issues scenarios people happen to have while working with k8s that are worth benchmarking against. this will make the time increase even more, currently:

> Total Time ~= sum(tasks)

The solution for the _"bottleneck"_ of executing all tasks sequentially would be using goroutines and channels to increase the throughput, where: 

> Total Time ~= longest individual task time + overhead

I also wondered if an hybrid approach would be best but I leave that open to your recommendations. 

Let me know if an hybrid approach allowing configurable concurrency, even down to sequential execution or a fully goroutine approach would be best, I would like to work on implementing this.

zvdy:
# K8s-bench Evaluation Results

## Model Performance Summary

| Model | Success | Fail |
|-------|---------|------|
| grok-3-beta | 9 | 1 |
| **Total** | 9 | 1 |

## Overall Summary

- Total Runs: 10
- Overall Success: 9 (90%)
- Overall Fail: 1 (10%)

## Model: grok-3-beta

| Task | Provider | Result |
|------|----------|--------|
| create-network-policy | grok | ❌  |
| create-pod | grok | ✅ success |
| create-pod-mount-configmaps | grok | ✅ success |
| create-pod-resources-limits | grok | ✅ success |
| fix-crashloop | grok | ✅ success |
| fix-image-pull | grok | ✅ success |
| fix-service-routing | grok | ✅ success |
| list-images-for-pods | grok | ✅ success |
| scale-deployment | grok | ✅ success |
| scale-down-deployment | grok | ✅ success |

**grok-3-beta Summary**

- Total: 10
- Success: 9 (90%)
- Fail: 1 (10%)

---

_Report generated on May 7, 2025 at 8:54 PM_

zvdy:
[...]

PR #199, opened 2025-05-09
k8s-bench concurrent task execution

zvdy:
# Motivation

As mentioned in #158, instead of executing tasks sequentially (where total time equals the sum of all task durations) as current implementation does, tasks now run in parallel using `goroutines` (where total time approaches the longest individual task plus minimal overhead).

## Features

- Add `--concurrency` flag to control parallelism:
  - 0 (default): Auto-configure to match the number of tasks
  - 1: Run sequentially (original behavior)
  - N: Run N tasks concurrently

- Dynamic adjustment based on available tasks
- Proper error handling and result collection from all workers
- Updated documentation in README.md

## Technical Implementation

The implementation uses Go's concurrency primitives:

1. **Worker Pool Pattern**:
   - Created a bounded worker pool with configurable size based on the concurrency flag
   - Each worker runs in its own goroutine and processes tasks from a shared channel

2. **Channel Coordination**:
   - `taskCh`: Distributes tasks to worker goroutines
   - `resultsCh`: Collects evaluation results from all workers
   - `errorsCh`: Captures errors that might occur during execution

3. **Synchronization**:
   - `sync.WaitGroup` ensures all workers complete before aggregating results
   - Channel capacity is pre-allocated based on the number of tasks to prevent blocking

4. **Resource Management**:
   - Each worker handles its own file I/O for logs and result files
   - Output directory structure remains consistent with the sequential implementation

This implementation maintains the existing functionality while providing significant performance improvements for multi-task benchmarks. Which can also be implemented into a testing suite or CI/CD pipeline with github actions #196 

### Setup all concurrent tasks:
image

image

### DIFF
--- a/k8s-bench/README.md
+++ b/k8s-bench/README.md
@@ -24,6 +24,9 @@ The `run` subcommand executes the benchmark evaluations.
 # Run evaluation for a specific LLM provider and model with tool use shim disabled
 ./k8s-bench run --llm-provider=grok --models=grok-3-beta --agent-bin ../kubectl-ai --task-pattern=fix-probes --enable-tool-use-shim=false --output-dir .build/k8sbench
 
+# Run evaluation sequentially (one task at a time)
+./k8s-bench run --agent-bin <path/to/kubectl-ai/binary> --tasks-dir ./tasks --output-dir .build/k8sbench --concurrency 1
+
 # Run evaluation with all available options
 ./k8s-bench run \
   --agent-bin <path/to/kubectl-ai/binary> \
@@ -34,6 +37,7 @@ The `run` subcommand executes the benchmark evaluations.
   --models gemini-2.5-pro-preview-03-25,gemini-1.5-pro-latest \
   --enable-tool-use-shim true \
   --quiet true \
+  --concurrency 0 \
   --output-dir .build/k8sbench
 ```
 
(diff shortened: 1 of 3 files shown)


mikebz:
Can concurrent execution lead to tests colliding?

droot:
> Can concurrent execution lead to tests colliding?

No because each task sets up resources in a dedicated namespace. Though, ensuring uniqueness of namespaces is left to the developers which is not hard today because we have only ~14-15 evals. But as we add more tasks, we can probably find a way to detect the collision or avoid it completely.

zvdy:
I assume by tests you mean the tasks themselves, they run on their own goroutine, so they execute exactly as before, all pass as expected and the reports are the same.

The only difference, is that the outputs of the model processing "floods" the terminal as it performs many tasks on multiple namespaces and analysis at the same time, but, as the tasks are not expected to be something interactive I don't think this is something to be worried about, also, you have the possibility to don't use concurrency, and run sequentially as it was done before

droot (approved):
This is big improvement. I was little skeptical of the default value for concurrency, but I want to see how it plays out in practice.

Thank you @zvdy

zvdy:
> > Can concurrent execution lead to tests colliding?
> 
> 
> 
> No because each task sets up resources in a dedicated namespace. Though, ensuring uniqueness of namespaces is left to the developers which is not hard today because we have only ~14-15 evals. But as we add more tasks, we can probably find a way to detect the collision or avoid it completely.

We can check if namespaces collide, but I assume no task should be as similar as other existing one to reuse the namespace, still, we can check the namespaces before setting up and running the tasks

PR #203, opened 2025-05-09
add file extensions for discovery within IDEs and find.

mikebz:
This pull request updates the CI presubmit workflows and associated scripts to ensure consistency in naming conventions and improve maintainability. The key changes involve renaming scripts to include a `.sh` extension and updating references accordingly in the CI workflow configuration.


mikebz:
I am guessing that changes to pre-submits somehow need to bypass the checking for those changes @justinsb @droot this one might have to skip the ci-presubmit checks.

justinsb:
So the idea of not putting an extension is that the `.sh` extension is an implementation detail.  These scripts could be written in python / go / perl / whatever in future.

That said it's not a blocker to add a .sh extension.  But in terms of upside: what IDE fails to recognize them as shell scripts?

mikebz:
image
correct the IDE doesn't see them as such until you open them up.

droot:
> So the idea of not putting an extension is that the .sh extension is an implementation detail. These scripts could be written in python / go / perl / whatever in future.

In the short term, I don't see these getting rewritten in other languages, so adding the extension sounds reasonable to me.

> That said it's not a blocker to add a .sh extension.

Sounds good. 


I will go ahead and get these changes in.

droot (approved):

PR #205, opened 2025-05-10
Add support for additional prompt template files

### LINKED ISSUE #204: Support additional prompt files
I'd like to be able to supply additional prompt files that will be combined with the default prompt file (not overwrite it). My use-case is to adjust the style of the responses.

RRethy:
Close #204 

Add `--additional-prompt-template-file-paths` flag with templates that are appended to the default prompt file when generating the prompt.

### DIFF
--- a/cmd/main.go
+++ b/cmd/main.go
@@ -82,13 +82,14 @@ type Options struct {
 	EnableToolUseShim bool `json:"enableToolUseShim,omitempty"`
 	// Quiet flag indicates if the agent should run in non-interactive mode.
 	// It requires a query to be provided as a positional argument.
-	Quiet                  bool   `json:"quiet,omitempty"`
-	MCPServer              bool   `json:"mcpServer,omitempty"`
-	MaxIterations          int    `json:"maxIterations,omitempty"`
-	KubeConfigPath         string `json:"kubeConfigPath,omitempty"`
-	PromptTemplateFilePath string `json:"promptTemplateFilePath,omitempty"`
-	TracePath              string `json:"tracePath,omitempty"`
-	RemoveWorkDir          bool   `json:"removeWorkDir,omitempty"`
+	Quiet                             bool     `json:"quiet,omitempty"`
+	MCPServer                         bool     `json:"mcpServer,omitempty"`
+	MaxIterations                     int      `json:"maxIterations,omitempty"`
+	KubeConfigPath                    string   `json:"kubeConfigPath,omitempty"`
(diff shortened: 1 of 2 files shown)


Mcvsguy56 (approved):

cmd/main.go
```diff
@@ -613,7 +617,7 @@ func resolveKubeConfigPath(opt *Options) error {
 
 func startMCPServer(ctx context.Context, opt Options) error {
 	workDir := filepath.Join(os.TempDir(), "kubectl-ai-mcp")
-	if err := os.MkdirAll(workDir, 0755); err != nil {
+	if err := os.MkdirAll(workDir, 0o755); err != nil {
```
droot, line 620: Can we preserve the original permissions ?

cmd/main.go
```diff
@@ -257,6 +259,7 @@ func (opt *Options) bindCLIFlagsToViper(f *pflag.FlagSet) error {
 	f.IntVar(&opt.MaxIterations, "max-iterations", opt.MaxIterations, "maximum number of iterations agent will try before giving up")
 	f.StringVar(&opt.KubeConfigPath, "kubeconfig", opt.KubeConfigPath, "path to kubeconfig file")
 	f.StringVar(&opt.PromptTemplateFilePath, "prompt-template-file-path", opt.PromptTemplateFilePath, "path to custom prompt template file")
+	f.StringArrayVar(&opt.AdditionalPromptTemplateFilePaths, "additional-prompt-template-file-paths", opt.AdditionalPromptTemplateFilePaths, "additional paths to custom prompt template files")
```
droot, line 262: The name is getting super verbose :) One option is to name it `extra-prompt-paths` (extra replaces additional pretty well)

Thinking out loud here: if the paths point to a directory, we can slurp all files from that directory (this could be added later as well doesn't have to be this PR).

PS: We will introduce short flags soon to help with the verbosity.

droot (changes requested):
Thanks for the PR, I have minor comments.

cmd/main.go
RRethy: Editor added that, sorry. Reverting.

RRethy:
Addressed comments. However, it's no longer a feature I require so feel free to close the PR if you don't want to add this feature yet.

droot:
> Addressed comments. However, it's no longer a feature I require so feel free to close the PR if you don't want to add this feature yet.

Good to know. What changed ?

droot (approved):
Getting this in because it will make it easy to augment the system prompt. Thanks @RRethy

PR #208, opened 2025-05-10
gollm: add `mock_gollm`  to mock LLM responses 

yumosx:
simulate LLM responses to test without real LLMs.

ref: #196


droot:
Thanks @yumosx for the PR. We are still assessing how and at what level we want to mock the components for the AI agent, so keeping the PR on hold for now. Would be great if you can share your experience of using this mocking framework.

yumosx:
I noticed that this feature is actually already supported, so I closed this PR.

PR #213, opened 2025-05-11
Add SHA-256 checksum verification to model download script

seunggihong:
### Summary

This PR enhances the 'Model Serving' download script by introducing SHA-256 checksum verification for the downloaded model file 'gemma-3-12b-it-Q4_K_M.gguf'.

### Changes

- Add 'EXPECTED_SHA256' as a configurable variable.
- If the expected hash is not set, the script outputs and exits the actual hash.
- When a hash is set, the script checks the downloaded file and quickly fails in case of a mismatch.
- If checksums do not match, delete invalid files.

### important reason

This verifies the integrity after download, improving the safety and reproducibility of the pre-trained model. It also makes onboarding and deployment robust by capturing corrupted or tampered files early.

### Use

1. Run the script once without setting 'EXPECTED_SHA256'.
2. Copy the printed hash and set it to 'EXPECTED_SHA256'.
3. Subsequent runs automatically verify the file.

Thank you for this awesome project.


modelserving/dev/tasks/download-model
```diff
@@ -10,8 +10,52 @@ cd "${SRC_DIR}"
...
+
+  echo "Checksum verified"
+}
+
+if [[ ! -f "${MODEL_PATH}" ]]; then
+  download_model
+  verify_checksum
+else
+  echo "${MODEL_NAME} already exists. Verifying checksum..."
+  verify_checksum
```
droot, line 60: This change would have been more useful if it also fetched the expected checksum from the original source.

seunggihong: Thanks for the feedback.

droot: Thank you!

I am curious what prompted you make this change (ofc integrity check is awesome :)) but want to understand if you ran into an issue with the tooling because of the integrity check not being there.

droot (approved):
The PR is all good. Just a comment about gathering more data.

modelserving/dev/tasks/download-model
seunggihong: Thanks for asking. 

I didn’t run into an actual issue yet, but this was more of a proactive measure.

PR #214, opened 2025-05-11
Remove viper integration

droot:
I think we added viper integration too soon. I would like to reconsider this decision and re-introduce it more deliberately when the project is ready to taken on the complexity (and config expansion) that viper integration brings.

Input to a CLI program (env-variables, commandline flags, ...etc.) are like an API and need to be maintained forever once out, so we need to be deliberate about every single input configuration we add. viper introduces multiple ways to configure the inputs and I am not so sure project need that at the moment. And I am finding that it increases the complexity and add dependencies to the project without any significant benefits.

I am sorry, I should have looked into it when we were integrating cobra, that's on me. /cc @selimacerbas @tuannvm


tuannvm:
No worry. As long as we have a straight-forward and conventional way to set flag, we're in good shape.

selimacerbas:
I think we can add it anytime anyway so I am also good with it. We can certainly prioritise the functionality of flags and make it good enough.

justinsb (approved):

justinsb:
I am glad we tried it out, and glad that we can iterate quickly when we find bumps in the road!  Thanks all and let's just keep moving forwards!

droot:
Thanks everyone, getting this in.

tuannvm:
Great, thanks @droot. Then I'll continue the on-going PRs with non-viper implementation.

PR #221, opened 2025-05-12
Feature: add shields in README

zvdy:
## Add Badges/Shields on README

It's a pretty common use case for OSS to have badges, they are convenient and make the repository be more _appealing_ if you could say so:

- `shields.io` uses [Creative Commons Zero](https://github.com/badges/shields/blob/master/LICENSE) License 

- `goreportcard` uses [Apache 2.0](https://github.com/gojp/goreportcard/blob/master/LICENSE) License.


droot (approved):
Thanks. Let's try it.

Issue #222, opened 2025-05-12
Support custom tools


janetkuo:
This allows users to extend `kubectl-ai` to leverage arbitrary tools defined locally, without adding everything as built-in tools. 

Related: https://github.com/GoogleCloudPlatform/kubectl-ai/issues/218

droot:
I am collecting the use-cases. can you share specific example you ran into ?

janetkuo:
For example, `nomos` CLI for Config Sync troubleshooting.

janetkuo:
Another example is that someone wants to troubleshoot their clusters with kubectl-ai + some internal tools.

PR #223, opened 2025-05-12
Support custom tools

janetkuo:
Fixes https://github.com/GoogleCloudPlatform/kubectl-ai/issues/222

Usage: `kubectl-ai --custom-tools-config=<path to tools config>` (default location `~/.config/kubectl-ai/tools.yaml`)

Example tool config (this takes multiple custom tools):
```yaml
- name: nomos_status
  description: "Runs the 'nomos status' command to check Config Sync status on specified Kubernetes clusters."
  command: "nomos status"
  command_desc: "The nomos status command for Config Sync status on the cluster"
```
(Tool description gives LLM hint when to use the tool; command description is more about syntax and examples of command usage.)

Example result:
Screenshot 2025-05-15 at 11 43 35 AM
Screenshot 2025-05-15 at 11 43 49 AM

Note that `kubectl-ai` can still answer questions without the tool. Adding custom tools enriches the answer. 
Screenshot 2025-05-15 at 2 59 17 PM

### DIFF
--- a/cmd/main.go
+++ b/cmd/main.go
@@ -86,6 +86,7 @@ type Options struct {
 	ExtraPromptPaths       []string `json:"extraPromptPaths,omitempty"`
 	TracePath              string   `json:"tracePath,omitempty"`
 	RemoveWorkDir          bool     `json:"removeWorkDir,omitempty"`
+	ToolConfigPath         string   `json:"toolConfigPath,omitempty"`
 
 	// UserInterface is the type of user interface to use.
 	UserInterface UserInterface `json:"userInterface,omitempty"`
@@ -236,6 +237,12 @@ func run(ctx context.Context) error {
 		return fmt.Errorf("failed to load config file: %w", err)
 	}
 
+	// Load and register custom tools from config file
+	if err := tools.LoadAndRegisterCustomTools(opt.ToolConfigPath); err != nil {
+		// Log the error but continue execution, as custom tools are optional
+		klog.Warningf("Failed to load or register custom tools (path: %q): %v", opt.ToolConfigPath, err)
+	}
+
 	rootCmd, err := BuildRootCommand(&opt)
 	if err != nil {
 		return err
@@ -267,6 +274,7 @@ func (opt *Options) bindCLIFlags(f *pflag.FlagSet) error {
 	f.StringVar(&opt.ModelID, "model", opt.ModelID, "language model e.g. gemini-2.0-flash-thinking-exp-01-21, gemini-2.0-flash")
 	f.BoolVar(&opt.SkipPermissions, "skip-permissions", opt.SkipPermissions, "(dangerous) skip asking for confirmation before executing kubectl commands that modify resources")
 	f.BoolVar(&opt.MCPServer, "mcp-server", opt.MCPServer, "run in MCP server mode")
+	f.StringVar(&opt.ToolConfigPath, "custom-tools-config", opt.ToolConfigPath, "path to custom tools config file")
 	f.BoolVar(&opt.EnableToolUseShim, "enable-tool-use-shim", opt.EnableToolUseShim, "enable tool use shim")
 	f.BoolVar(&opt.Quiet, "quiet", opt.Quiet, "run in non-interactive mode, requires a query to be provided as a positional argument")
 
--- a/pkg/tools/custom_tool.go
+++ b/pkg/tools/custom_tool.go
@@ -1,0 +1,96 @@
+// Copyright 2025 Google LLC
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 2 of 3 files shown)


go.mod
```diff
@@ -15,6 +15,7 @@ require (
 	github.com/mark3labs/mcp-go v0.26.0
 	github.com/spf13/cobra v1.9.1
 	github.com/spf13/pflag v1.0.6
+	gopkg.in/yaml.v3 v3.0.1
```
droot, line 18: I think this package is archived, can we use sigs.k8s.io/yaml ? that's already included.

cmd/main.go
```diff
@@ -228,6 +228,12 @@ func run(ctx context.Context) error {
 		return fmt.Errorf("failed to load config file: %w", err)
 	}
 
+	// Load and register custom tools from config file
```
droot, line 231: nit: would be nice to take a path from the command-line (`--extra-tools-conf` or `--custom-tools-conf` or something) and default to ~/.config/kubectl-ai/tools.yaml

pkg/tools/custom_tool.go
```diff
@@ -0,0 +1,81 @@
...
+
+	"github.com/GoogleCloudPlatform/kubectl-ai/gollm"
+)
+
+// CustomToolConfig defines the structure for configuring a custom tool.
+type CustomToolConfig struct {
+	Name    string                    `yaml:"name"`
+	FnDef   *gollm.FunctionDefinition `yaml:"fndef"`
+	Command []string                  `yaml:"command"` // Command and args to execute
+}
```
droot, line 31: would be great if you can share an example of what a real config look like say for `nomos` or `helm`

Better will be a quick screencast showing this in action to get a feel for the UX.

droot (changes requested):
Thanks Janet for taking a stab at it, I have some minor comments. Would it be possible to share a screencast or examples of config/tool-run to see the UX.

go.mod
janetkuo: Good catch. Fixed.

pkg/tools/custom_tool.go
janetkuo: Sure, will add an example or demo once I've tested this out.

pkg/tools/tools.go
```diff
@@ -180,3 +183,49 @@ func ToolResultToMap(result any) (map[string]any, error) {
...
+
+// LoadAndRegisterCustomTools loads tool configurations from a YAML file
+// and registers them.
+func LoadAndRegisterCustomTools() error {
+	home, err := os.UserHomeDir()
+	if err != nil {
+		return fmt.Errorf("failed to get user home directory: %w", err)
+	}
+	// Custom tools are stored in ~/.config/kubectl-ai/tools.yaml
+	configPath := filepath.Join(home, ".config", "kubectl-ai", "tools.yaml")
```
hakman, line 195: This has the potential to become quite large. Could each tool be a separate file in a tools directory like `$HOME/.config/kubectl-ai/tools/helm.yaml`?

cmd/main.go
hakman: Could also be `--tools-dir` and accept multiple values. This would make it possible to extract the current tools to a `tools` dir inside the repo/install dir, look into some well-known locations like `$HOME/.config/kubectl-ai/tools/` and also also specify additional locations.

pkg/tools/custom_tool.go
janetkuo: @droot Updated the example config and screenshot in https://github.com/GoogleCloudPlatform/kubectl-ai/pull/223#issue-3058426043

cmd/main.go
janetkuo: I've updated the example config in https://github.com/GoogleCloudPlatform/kubectl-ai/pull/223#issue-3058426043. I prefer a single config file for all tools given that having to create a file per tool is more work.

pkg/tools/custom_tool.go
janetkuo: @droot Do we need both the tool description and command description?

droot: Awesome. Taking a look soon.

cmd/main.go
droot: I was thinking about the UX a bit more, and I realized supporting the single-yaml-per-tool has an advantage. One can simply share the helm.yaml in the instructions in README or even host it do wget/curl it. Even better, point kubectl-ai to a commandline tool, it can run `--help` and generate the tool.yaml. 

An example:
```
kubectl-ai tool-manifest <path/to/the/tool>
Saved nomos.yaml to ~/.kubectl-ai/tools/nomos.yaml
```
We can probably support both the format (tools.yaml and single-file-per-tool.yaml) with little bit of pain in the code :).

I think we can our tool parsing logic can detect if it is a collection of tools or single tool, so probably not a blocker to get this in first and later add support for single-yaml-per-tool. WDYT ?

/cc @hakman

pkg/tools/tools.go
droot: I discussed this in a comment thread. pl. take a look. My feedback is driven more by the UX.

cmd/main.go
janetkuo: I like the idea of generating custom tool configs using LLM. I think both separate files and folders are legit use cases, and we could eventually support both. I will keep it a single file as the initial change.

droot:
ok, lets get this in. We can tweak this in follow ups. You will have to rebase it.

PR #230, opened 2025-05-14
fix(tools): Use exec.LookPath to find bash executable before running bash commands

### LINKED ISSUE #228: [BUG] error occurs when running kubectl command (tool) in NixOS
In NixOS, there is no `/bin/bash` or `/usr/bin/bash`, and the command tries to execute command through bash causes error:

```log
action: tool-response
payload:
  error: 'fork/exec /bin/bash: no such file or directory'
  id: 4d78398b-b669-41f4-bc3e-b0d58df6ff0f
  response: null
timestamp: "2025-05-14T14:51:02.994004384+08:00"
```

ruinshe:
Fixes #228

Find the bash executable using `exec.LookPath` instead of hardcoding `/bin/bash`. Falls back to `/bin/bash` if not found in PATH. Improves compatibility on systems where bash might be installed in a non-standard location.

On some systems (e.g. NixOS), executables might not be in standard locations like /bin/bash.

### DIFF
--- a/pkg/tools/bash_tool.go
+++ b/pkg/tools/bash_tool.go
@@ -24,15 +24,24 @@ import (
 	"strings"
 
 	"github.com/GoogleCloudPlatform/kubectl-ai/gollm"
+	"k8s.io/klog/v2"
 )
+
+var bashBin string
 
 func init() {
+	// Find the bash executable path using exec.LookPath.
+	// On some systems (like NixOS), executables might not be in standard locations like /bin/bash.
+	var err error
+	bashBin, err = exec.LookPath("bash")
+	if err != nil {
+		log := klog.FromContext(context.Background())
+		log.Info("Warning: 'bash' not found in PATH, defaulting to /bin/bash", "error", err)
+		bashBin = "/bin/bash"
+	}
+
 	RegisterTool(&BashTool{})
 }
-
-const (
-	bashBin = "/bin/bash"
-)
 
 // expandShellVar expands shell variables and syntax using bash
 func expandShellVar(value string) (string, error) {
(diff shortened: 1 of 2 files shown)


ruinshe:
Currently I modified the `bashBin` in bash_tool.go in init function directly, if you think it's not a good choice and prefer passing the bash command path using parameters, or using other ways, please that me know.

pkg/tools/bash_tool.go
```diff
@@ -24,16 +24,25 @@ import (
...
+	"k8s.io/klog/v2"
 )
 
+var bashBin string
+
 func init() {
+	// Find the bash executable path using exec.LookPath.
+	// On some systems (like NixOS), executables might not be in standard locations like /bin/bash.
+	var err error
+	bashBin, err = exec.LookPath("bash")
```
droot, line 36: Can we move this lookup closer to the command execution. Keeping package initialization side-effect free is preferable and also in some cases (simple question/answer) bash is never even invoked.

Also, kubectl execution is also done using bash, so lets do it in a way it works for both.

droot (changes requested):
Thank you! just a minor comment.

droot:
> Currently I modified the `bashBin` in bash_tool.go in init function directly, if you think it's not a good choice and prefer passing the bash command path using parameters, or using other ways, please that me know.

Would prefer to keep init() side-effect free, suggested to move it closer to command construction. and its okay to duplicate code if needed.

pkg/tools/bash_tool.go
ruinshe: Fixed by adding a new function `lookupBashBin`, please take a look, thanks!

droot:
Thanks @ruinshe  you will need to rebase the PR because we had to update the presubmits, sorry :(

ruinshe:
> Thanks @ruinshe you will need to rebase the PR because we had to update the presubmits, sorry :(

done

droot (approved):

PR #231, opened 2025-05-14
Add Krew release automation

LefterisXefteris:


droot:
Thanks for the PR. I have one question about the dependency on our release workflow. 

We are using goreleaser to publish releases on very new tag and I am assuming release needs to be published before we publish the krew manifest. I also remember seeing goreleaser itself providing a mechanism to integrate krew automation ( I will have to look into it).

droot:
@LefterisXefteris gentle reminder about the question above :)

LefterisXefteris:
Hi @droot,

You’re exactly right—the GitHub Release needs to exist before we bump the Krew manifest. In our workflow, the goreleaser-action step runs first (publishing the release for the new v* tag), and only once that completes do we invoke krew-release-bot.

If you’d rather consolidate into GoReleaser’s built-in Krew support, we can also add a krew: block to our .goreleaser.yml (see https://goreleaser.com/plugins/krew/) so that GoReleaser itself handles the manifest PR after publishing.

Let me know which approach you’d prefer and I’ll update the PR accordingly!

Thanks,
Lefteris

LefterisXefteris:
Hi @droot,
Thanks for the reminder! To confirm: would you prefer that we drop the standalone krew-release-bot step and instead rely on GoReleaser’s built-in krews publisher to automatically open the PR against kubernetes-sigs/krew-index after tagging?

Please let me know if that approach works—or if you’d rather keep the existing bot step. Thanks!

droot:
@LefterisXefteris 

Lets add the `krew-release-bot` step in the existing [release workflow](https://github.com/GoogleCloudPlatform/kubectl-ai/blob/main/.github/workflows/release.yaml) as shown in the example here https://github.com/rajatjindal/krew-release-bot?tab=readme-ov-file#example-when-using-go-releaser (ofc use the action image that you have).

This will ensure that the pull request for krew-index is created after we have published the release and will also consolidate all the release steps in one workflow.

Thank you.

droot:
`krew` automation is now place, so we can close this out.

PR #233, opened 2025-05-14
Lazy initialization of input readers and evals

droot:
Two separate but slightly related changes (captured in separate commits)

## Lazy initializations of Input readers (tty and readline)

Initialize these readers only when needed. Two benefits:
1. We won't setup these when running in non-interactive mode (`--quiet`)
2. k8s-bench feeds the interactive script on stdin, that forces the kubectl-ai to use tty for input when the input is through stdin. And github actions don't make /dev/tty available and that causes `k8s-bench` to fail when running in CI (github actions)

CI changes
1. Added step to run lightweight evals (two tasks related to scaling)
2. Added a workflow for periodic evaluation

Note: I am using keyless setup for authenticating github actions with GCP's vertexai platform. I expect a few bumps. I have tested this in my own fork, but let see how it goes here. 

*UPDATE*: Had to remove the presubmit checks because I am running into permission issue. Will introduce that as separate PR.

### DIFF
--- a/pkg/ui/terminal.go
+++ b/pkg/ui/terminal.go
@@ -37,9 +37,9 @@ type TerminalUI struct {
 	subscription io.Closer
 
 	// Input handling fields (initialized once)
-	rlInstance *readline.Instance // For readline input
-	ttyFile    *os.File           // For TTY input
-	ttyReader  *bufio.Reader      // For TTY input
+	rlInstance        *readline.Instance // For readline input
+	ttyFile           *os.File           // For TTY input
+	ttyReaderInstance *bufio.Reader      // For TTY input
 
 	// currentBlock is the block we are rendering
 	currentBlock Block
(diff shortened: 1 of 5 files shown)


pkg/ui/terminal.go
```diff
@@ -70,42 +70,51 @@ func NewTerminalUI(doc *Document, journal journal.Recorder, useTTYForInput bool)
...
+		Stdout:      os.Stdout,
+		Stderr:      os.Stderr,
+		HistoryFile: historyPath,
+		// History enabled by default
+	})
+	if err != nil {
+		// Log warning or fallback if readline init fails?
+		klog.Warningf("Failed to initialize readline, input might be limited: %v", err)
+		// Proceed without readline for now, or return error?
+		// Returning error to make it explicit
```
justinsb, line 111: Yes, I think returning an error is correct at least until we figure out when this might fail.  If/when we find that, we might want an option to turn off readline.

justinsb (approved):

PR #241, opened 2025-05-16
fix(ci): Ensure job runs only for the correct repository

mikebz:
This pull request makes a small change to the `.github/workflows/ci-periodic.yaml` file by adding a condition to the `run-eval` job. The job will now only run if the repository is `GoogleCloudPlatform/kubectl-ai`.


droot (approved):
Thank you so much!

Issue #145, opened 2025-05-06
Help needed: We need more evals


droot:
`kubectl-ai` has 10 eval tasks today covering different areas. We want to improve the evals coverage. The goal is to get the eval coverage closer to realistic scenario that users run into.

This is an area where we need input from the community. Share the scenarios that you or your team run into very regularly.

Adding an eval task to `k8s-bench` is very simple, you can see existing tasks.

Ask:

- Share a scenario that we can add eval for
- Better, send a PR with the eval task following an example under `k8s-bench` directory.

/cc @selimacerbas @rxinui @cwrau @hakman @mattn @DoctorLai @mschneider82 @tuannvm

tuannvm:
@droot I can take on evals for openai

tuannvm:
It would be awesome if we could also add these features:

- Automatically run an evaluation on GitHub Actions whenever new models are added, released, or updated.
- Display the evaluation results back on k8s-bench.md.
- Use GitHub Actions with Kind to run a Kubernetes cluster in the CI environment.

Br1an67:
I'd like to work on this. I added a new eval task `fix-failed-job` under `k8s-bench/` to test AI debugging capabilities by simulating a failed Kubernetes Job caused by a command typo.
PR: #639

PR #253, opened 2025-05-20
eval: multi container pod communication with sidecar

snewcomer:
ref https://github.com/GoogleCloudPlatform/kubectl-ai/issues/145

This PR tests the following multi pod with volume sharing.  Similar to...

```
cat <<EOF | kubectl apply -f -
apiVersion: v1
kind: Pod
metadata:
  name: communication-pod
  namespace: multi-container-test
spec:
  volumes:
  - name: logs-volume
    emptyDir: {}
  
  containers:
  - name: web
    image: nginx:alpine
    ports:
    - containerPort: 80
    readinessProbe:
      tcpSocket:
        port: 80
      initialDelaySeconds: 5
      periodSeconds: 10
    volumeMounts:
    - name: logs-volume
      mountPath: /var/log/nginx
    - name: config-volume
      mountPath: /etc/shared-data
  
  - name: logger
    image: busybox
    command: ["tail", "-f", "/var/log/nginx/access.log"]
    volumeMounts:
    - name: logs-volume
      mountPath: /var/log/nginx
    startupProbe:
      exec:
        command:
        - sh
        - -c
        - "wget -q --spider http://localhost:80"
      initialDelaySeconds: 10
      periodSeconds: 5
      failureThreshold: 30
EOF
```


k8s-bench/tasks/multi-container-pod-communication/task.yaml
```diff
@@ -0,0 +1,7 @@
+script:
+- prompt: |
+    Create a pod named 'communication-pod' in the 'multi-container-test' namespace with two containers: 1. A 'web' container using nginx:alpine image that serves web content on port 80. 2. A 'logger' container using busybox image that monitors the nginx access logs. The containers should share a volume named 'logs-volume' where nginx writes its access logs. The web container should be ready when port 80 is available. The logger container should start only after the web container is ready.
```
droot, line 3: Two things:
1. Can we express this scenario to be a more realistic use-case. 
2. Think of it this way: we want to describe the intent of "what" we want to see in the cluster. We don't want to describe the exact steps to reach that desired outcome. The eval is a sort of a test for the AI model to come up with the right steps given a desired outcome. Think of a declarative expression of the intent.

droot (changes requested):

k8s-bench/tasks/multi-container-pod-communication/task.yaml
snewcomer: Great feedback.  Right a bit rougher.  How does this sound?  (latest commit)

```
In the multi-container-test namespace, run a pod called communication-pod with two containers: a web server that serves traffic and writes logs, and a logger that processes those logs from a shared volume. The logger should only start once the web server is ready. The pod should be considered ready when the web server is serving traffic.
```

droot: This is definitely looking better. Thanks @snewcomer.

For fun, I tried Gemini 2.5 pro for this task :)

```
Option 1 (Concise & Declarative):

Deploy a web application in the multi-container-test namespace. The application requires a dedicated logging container that processes the web server's logs from a shared disk. Ensure the logging container only starts after the web server is fully operational and serving requests. The entire application deployment should be marked as ready once the web server is live and accessible.

Option 2 (Slightly More Scenario-Focused):

I need to deploy a new web service, communication-pod, into our multi-container-test environment. This service consists of a primary web server that will generate access and error logs. To keep the web server performant, a separate container must handle the processing of these logs, which will be shared via a volume. It's crucial that the log processing only begins once the web server is confirmed to be up and actively serving traffic. The overall deployment is considered successful and ready when users can reach the web server.

Option 3 (User Story Format):

As a developer, I want to deploy my web application, communication-pod, to the multi-container-test namespace. This application has two main parts: the web server itself and a log processor. The web server writes its logs to a shared volume, and the log processor reads from there. To ensure proper log ingestion, the log processor container must wait until the web server container is fully initialized and actively handling requests before it starts. The deployment should signal readiness only when the web server is successfully serving traffic.
```

Evals are very important to the project, so thank you for contributing to it. We published roadmap yesterday https://github.com/GoogleCloudPlatform/kubectl-ai/issues/265

would love your feedback.

droot (approved):

PR #256, opened 2025-05-20
Feat/#252 version handler

### LINKED ISSUE #252: Bug: Cannot get version without GEMINI_API_KEY
We have a script for our developers that will set up all dependencies with a specific version, but we can't even get the version of the CLI without providing a GEMINI_API_KEY. I would say for that no api key should be required.
Our current workaround is just setting it to anything:
`GEMINI_API_KEY=notfound kubectl ai version --quiet`

zvdy:
### Summary

- Added a new `version` command to the CLI outside of the repl. Now you can run `kubectl-ai version` to see the current version, commit, and build date of the tool without the need to provide env variables. - Updated our release process so that every build now includes this version information automatically. This is set up in our GoReleaser configuration. previous [Releases](https://github.com/GoogleCloudPlatform/kubectl-ai/actions/workflows/release.yaml) still reflected default config:

```go
var (
	version = "dev"
	commit  = "none"
	date    = "unknown"
)
```

### Tests

- Built the project using GoReleaser locally in snapshot mode to simulate a mock a _CI/CD Release_ 
https://github.com/GoogleCloudPlatform/kubectl-ai/blob/b8e4c66e27f3235e0999c1d908eb98bae448b700/.github/workflows/release.yaml#L26-L33

- Extracted the built binary from the dist folder.
- Ran `./kubectl-ai version` and confirmed it prints out the correct version, commit, and date, like this:

 ```
❯ dist/kubectl-ai version
kubectl-ai version: 0.0.0-SNAPSHOT-b8e4c66
commit: b8e4c66e27f3235e0999c1d908eb98bae448b700
date: 2025-05-20T17:07:50Z
  ```

> [!WARNING]
> Until a Release is released, the state of the variable will remain as the zero/current value:

```
❯ ./kubectl-ai version
kubectl-ai version: dev
commit: none
date: unknown
```

### DIFF
--- a/.goreleaser.yaml
+++ b/.goreleaser.yaml
@@ -23,6 +23,11 @@ builds:
       - windows
       - darwin
     main: ./cmd
+    ldflags:
+      - -s -w
+      - -X main.version={{.Version}}
+      - -X main.commit={{.Commit}}
+      - -X main.date={{.Date}}
 
 archives:
   - formats: [tar.gz]
--- a/cmd/main.go
+++ b/cmd/main.go
@@ -490,7 +490,7 @@ func (s *session) repl(ctx context.Context, initialQuery string) error {
 		case query == "clear":
 			s.ui.ClearScreen()
 		case query == "exit" || query == "quit":
-			// s.ui.RenderOutput(ctx, "Allright...bye.\n")
+			// s.ui.RenderOutput(ctx, "Alright...bye.\n")
 			return nil
 		default:
 			if err := s.answerQuery(ctx, query); err != nil {
--- a/cmd/main.go
+++ b/cmd/main.go
@@ -63,7 +63,7 @@ func BuildRootCommand(opt *Options) (*cobra.Command, error) {
 		Use:   "version",
 		Short: "Print the version number of kubectl-ai",
 		Run: func(cmd *cobra.Command, args []string) {
-			fmt.Printf("kubectl-ai version: %s\ncommit: %s\ndate: %s\n", version, commit, date)
+			fmt.Printf("version: %s\ncommit: %s\ndate: %s\n", version, commit, date)
 			os.Exit(0)
 		},
 	})
--- a/cmd/main.go
+++ b/cmd/main.go
@@ -59,6 +59,15 @@ func BuildRootCommand(opt *Options) (*cobra.Command, error) {
 		},
 	}
 
+	rootCmd.AddCommand(&cobra.Command{
+		Use:   "version",
+		Short: "Print the version number of kubectl-ai",
+		Run: func(cmd *cobra.Command, args []string) {
+			fmt.Printf("kubectl-ai version: %s\ncommit: %s\ndate: %s\n", version, commit, date)
+			os.Exit(0)
+		},
+	})
+
 	if err := opt.bindCLIFlags(rootCmd.Flags()); err != nil {
 		return nil, err
 	}


.goreleaser.yaml
```diff
@@ -23,6 +23,11 @@ builds:
       - windows
       - darwin
     main: ./cmd
+    ldflags:
+      - -s -w
+      - -X main.version={{.Version}}
+      - -X main.commit={{.Commit}}
+      - -X main.date={{.Date}}
```
droot, line 30: goreleaser does it by default: https://goreleaser.com/cookbooks/using-main.version/

That's the reason, all our released versions has correct version/commit/date set.

So this change isn't necessary (though it doesn't hurt).

cmd/main.go
```diff
@@ -59,6 +59,15 @@ func BuildRootCommand(opt *Options) (*cobra.Command, error) {
 		},
 	}
 
+	rootCmd.AddCommand(&cobra.Command{
+		Use:   "version",
+		Short: "Print the version number of kubectl-ai",
+		Run: func(cmd *cobra.Command, args []string) {
+			fmt.Printf("kubectl-ai version: %s\ncommit: %s\ndate: %s\n", version, commit, date)
```
droot, line 66: nits: can we remove kubectl-ai from the output

zvdy:
> Note: I did some testing in my main branch,  force pushed to remove the previous commit,

 @droot we can use the flag `--parallelism int` to speed up the release process 

image

Let me know if you want me to add it in this PR ttoo

droot (approved):
Thank you!

droot:
> @droot we can use the flag `--parallelism int` to speed up the release process

Oh, I thought it will defaults to available cores because we don't even know what does github action free tier gives us.

Separate PR is better.

.goreleaser.yaml
zvdy: yeah just realized this during my test, if i release locally, without the cfg, I get:

```
❯ tar -xzf dist/kubectl-ai_Linux_x86_64.tar.gz -C dist/ && ./dist/kubectl-ai version
version: dev
commit: none
date: unknown
```

but with them

```
❯ tar -xzf dist/kubectl-ai_Linux_x86_64.tar.gz -C dist/ && ./dist/kubectl-ai version
version: 1.0.0-SNAPSHOT-1ea95cf
commit: 1ea95cf5ff49e8eb24c102573844f5cee649fbeb
date: 2025-05-20T18:59:53Z
```

Just tried building from tag0.0.9 and those vars are set by default 

```
❯ ./kubectl-ai version

  Version:  0.0.9  
```

zvdy:
> > @droot we can use the flag `--parallelism int` to speed up the release process
> 
> Oh, I thought it will defaults to available cores because we don't even know what does github action free tier gives us.
> 
> Separate PR is better.

Yeah, defaults to 2vCPUs as ubuntu latest runner is set, my idea was to 

1. Utilize GKE as a Runner to improve build times

But the effort is not worth the _bottleneck_ at least for now in my opinion.

droot:
> But the effort is not worth the bottleneck at least for now in my opinion.

agreed. Release is fine with 2-3 minutes.

> Utilize GKE as a Runner to improve build times

This approach is useful for running evals. 


@zvdy was wondering how much time do you have to allocate to the project ? The reason being -- I am drafting roadmap items and was wondering if you are available to tackle some of those. (ofc no pressure :))

zvdy:
> @zvdy was wondering how much time do you have to allocate to the project ? The reason being -- I am drafting roadmap items and was wondering if you are available to tackle some of those. (ofc no pressure :)) 

Would love to work on it! I really enjoy collaborating on the project and having a roadmap and structure to work on would be really positive for the overall project

I assume this will be public perhaps using GitHub projects to track the work, but I might be wrong, if it's internal feel free to email me directly as needed :-)

droot:
> > @zvdy was wondering how much time do you have to allocate to the project ? The reason being -- I am drafting roadmap items and was wondering if you are available to tackle some of those. (ofc no pressure :))
> 
> Would love to work on it! I really enjoy collaborating on the project and having a roadmap and structure to work on would be really positive for the overall project
> 
> I assume this will be public perhaps using GitHub projects to track the work, but I might be wrong, if it's internal feel free to email me directly as needed :-)

excellent. We have been little adhoc in scheduling items to tackle, but we can do better in organizing the project and that has become priority.

And yes, the roadmap will be public. We will keep the process lightweight.

zvdy:
Sounds great, I will keep an eye out on this

Issue #146, opened 2025-05-06
POC request: Add mcp-client to enhance tool coverage


droot:
I think we can add `mcp-client` to extend coverage for tools that `kubectl-ai` can use. It will be cool project to explore what this integration might look like. We have rough idea, but POC is a good way to determine shape.

Any mcp enthusiast wants to take this up ?

/cc @tuannvm @selimacerbas

selimacerbas:
hey @droot thanks for the ping, i am willing to help on this. What kind of POC idea do you have in your mind?

droot:
Main goals of the POC are:

 - Investigate Go packages that we can use for mcp-client integration
 - How users will configure different MCP servers
 - Discovering tools exposed by the configured mcp servers and plug them into agentic loop
 - Discover key challenges/problems to be solved

Potential scenarios:
 - Fetch all deployments with unhealthy pods and send email with detailed report of each failed pod
 - Fetch logs (and YAML snapshots) of a particular deployment and save it to google drive
 - Fetch all apps updated in last 24 hrs and send health summary email
 - Given an app, create an SVG with diagrams for each associated k8s resource

tuannvm:
Thanks @droot for the ping. My common use case is to discover connection information and bridge it with MCP servers for further exploration. For example, identifying which Kafka brokers the pods are communicating with and then exploring further using [kafka-mcp-server](https://github.com/tuannvm/kafka-mcp-server).

My experience so far has been implementing the MCP server and client using [mcp-go](https://github.com/mark3labs/mcp-go). What are your thoughts?

droot:
> My experience so far has been implementing the MCP server and client using [mcp-go](https://github.com/mark3labs/mcp-go). What are your thoughts?

Not many people know, but `kubectl-ai` also has mcp server support for running kubectl commands. We used `mcp-go` and found it be excellent. I think Go team is releasing official mcp SDK which is based on the work of mcp-go. @selimacerbas I would suggest use mcp-go for the POC and save some time on the exploration.

> My common use case is to discover connection information and bridge it with MCP servers for further exploration. For example, identifying which Kafka brokers the pods are communicating with and then exploring further using [kafka-mcp-server](https://github.com/tuannvm/kafka-mcp-server).

can you say more about it. A concrete example with the exploration scenario.

selimacerbas:
[...]

PR #258, opened 2025-05-20
Feat/gdrive mcp-client(PoC)

selimacerbas:
Support gdrive to upload logs or any yaml outputs. mentioned in #146


selimacerbas:
@droot It is totally fine. I think I misunderstood the requirements but now it made sense after your explanation. So in this case, since gdrive does not have CLI, it is not suitable to use it as a extended custom tool because of the schema which expects a cli command. And understandably, we wouldn't want to cover all available services as built-in tool which would increase complexity. So to be able to extend non CLI services or any service, MCP Client comes into play. 

I will correctly implement mcp-client in my next PR.

droot:
@selimacerbas I see a draft PR from @tuannvm with `mcp-client` integration, worth taking a look https://github.com/GoogleCloudPlatform/kubectl-ai/pull/270

selimacerbas:
Thanks for letting me know @droot.

selimacerbas:
@droot, I assume that we would not want to have built-in gdrive tool. And mcp-client is on the way thanks to @tuannvm. Should I close this PR?

droot:
> @droot, I assume that we would not want to have built-in gdrive tool. And mcp-client is on the way thanks to @tuannvm. Should I close this PR?

Yes, we can close this PR. Thank you!

PR #261, opened 2025-05-21
gollm: bump default model for ollama provider

ameukam:
Bump to Gemma 3: https://ollama.com/library/gemma3

### DIFF
--- a/gollm/ollama.go
+++ b/gollm/ollama.go
@@ -37,7 +37,7 @@ func ollamaFactory(ctx context.Context, opts ClientOptions) (Client, error) {
 }
 
 const (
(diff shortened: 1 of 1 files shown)


gollm/ollama.go
```diff
@@ -37,7 +37,7 @@ func ollamaFactory(ctx context.Context, opts ClientOptions) (Client, error) {
 }
 
 const (
-	defaultOllamaModel = "gemma2:latest"
+	defaultOllamaModel = "gemma3:latest"
```
droot, line 40: How has been your experience of running it with gemma3 ?

Would be interested in running `k8s-bench` for gemma3 and share the results ?

droot (approved):

PR #270, opened 2025-05-22
feat(mcp): introduce MCP client with server management and tool exec

tuannvm:


cmd/main.go
```diff
@@ -71,7 +71,6 @@ func BuildRootCommand(opt *Options) (*cobra.Command, error) {
 	if err := opt.bindCLIFlags(rootCmd.Flags()); err != nil {
 		return nil, err
 	}
-
```
hakman, line 74: I think this should not be in the PR. I guess there's something missing from the PR?

go.mod
```diff
@@ -15,6 +15,7 @@ require (
 	github.com/mark3labs/mcp-go v0.26.0
 	github.com/spf13/cobra v1.9.1
 	github.com/spf13/pflag v1.0.6
+	gopkg.in/yaml.v3 v3.0.1
```
hakman, line 18: Can you use `sigs.k8s.io/yaml` instead?

hakman (changes requested):

go.mod
janetkuo: +1 this is deprecated

tuannvm:
@droot @janetkuo @hakman Could you help check if this looks good?

kubectl-ai mcp-client

README.md
```diff
@@ -201,6 +201,51 @@ A custom tool definition for `helm` could look like the following example:
     Use `helm --help` or `helm <subcommand> --help` to see full syntax, available flags, and examples for each command.
 ```
 
+## MCP Client Mode
+
+`kubectl-ai` can act as an MCP (Model Context Protocol) client, connecting to external MCP servers to access additional tools and capabilities beyond the built-in ones.
```
droot, line 206: I think we can simply say:

`kubectl-ai` can connect to external [MCP](http://mcp...) Servers to access additional tools in addition to built-in tools.

I think at some point, we will move them under one section `Using other tools` section

README.md
```diff
@@ -201,6 +201,51 @@ A custom tool definition for `helm` could look like the following example:
...
+      "args": ["-y", "@modelcontextprotocol/server-sequential-thinking"]
+    },
+    {
+      "name": "custom-server",
+      "command": "/path/to/your/mcp-server",
+      "args": ["--arg1", "value1"]
+    }
+  ]
+}
+```
```
droot, line 238: I am wondering if we should go with YAML syntax as other configuration files we have are in yaml. I think some of the existing tools (claude etc.) picked the JSON format because most of them were written in typescript and json is preferred format in that ecosystem. Since most of our users are k8s users, wrangling YAMLs in and out, we can prefer to keep it YAML.

cmd/main.go
```diff
@@ -300,9 +302,20 @@ func RunRootCommand(ctx context.Context, opt Options, args []string) error {
...
 			return fmt.Errorf("failed to start MCP server: %w", err)
 		}
+		return nil // MCP server mode blocks, so we return here
+	}
+
+	// Initialize MCP client if requested
+	if opt.MCPClient {
+		if err := tools.InitializeMCPClient(); err != nil {
+			klog.Warningf("Failed to initialize MCP client: %v", err)
+			// Continue with kubectl functionality - MCP failure should not block core features
```
droot, line 315: if mcp configuration exists, and we fail, then we should fail early to let the users know.

cmd/main.go
```diff
@@ -375,6 +388,16 @@ func RunRootCommand(ctx context.Context, opt Options, args []string) error {
...
 
+	// Show MCP server status in the welcome message only when MCP client is enabled
+	if opt.MCPClient {
+		if mcpBlocks, err := GetMCPServerStatusWithClientMode(opt.MCPClient); err == nil && len(mcpBlocks) > 0 {
+			doc.AddBlock(ui.NewAgentTextBlock().WithText("\nMCP Server Status:"))
+			for _, block := range mcpBlocks {
+				doc.AddBlock(block)
+			}
+		}
+	}
```
droot, line 399: Good call. I like it.

droot: we should also log the status in our log file.

pkg/tools/mcp_tool.go
```diff
@@ -0,0 +1,301 @@
...
+	}, nil
+}
+
+// =============================================================================
+// MCP Tool Implementation
+// =============================================================================
+
+// init function is kept minimal - MCP initialization is handled by --mcp-client flag
+func init() {
+	// MCP client initialization will be handled via explicit --mcp-client flag
```
droot, line 58: good call.

pkg/agent/systemprompt_template_default.txt
```diff
@@ -1,5 +1,10 @@
 You are `kubectl-ai`, an AI assistant with expertise in operating and performing actions against a kubernetes cluster. Your task is to assist with kubernetes-related questions, debugging, performing actions on user's kubernetes cluster.
 
+## MCP Server Status
+{{.MCPStatus}}
+
+You have access to tools from both local capabilities and connected MCP (Model Context Protocol) servers. When you use a tool from an MCP server, it will be clearly indicated with "[MCP]" in the execution log.
+
```
droot, line 7: I am curious what is the motivation behind this.

As far as the LLMs are concerned, there isn't any difference between the tools built-in, custom or mcp, so why highlight this ?

Prompt is an area where we should add stuff only when needed.

pkg/agent/systemprompt_template_default.txt
```diff
@@ -43,10 +48,13 @@ If you have enough information to answer the query:
 
 
 ## Remember:
+- You have access to both local tools (bash, kubectl, trivy) and MCP server tools (if connected).
+- MCP tools extend your capabilities with additional external services and providers.
```
droot, line 52: see my comment above.

pkg/agent/systemprompt_template_default.txt
```diff
@@ -43,10 +48,13 @@ If you have enough information to answer the query:
...
 - Fetch current state of kubernetes resources relevant to user's query.
 - Prefer the tool usage that does not require any interactive input.
 - For creating new resources, try to create the resource using the tools available. DO NOT ask the user to create the resource.
 - Use tools when you need more information. Do not respond with the instructions on how to use the tools or what commands to run, instead just use the tool.
+- If asked about your capabilities, mention both local tools and any connected MCP servers.
```
droot, line 57: see my comment above.

cmd/mcp_client.go
```diff
@@ -0,0 +1,252 @@
...
+		} else {
+			serverText += " - Not connected (--mcp-client disabled)"
+		}
+
+		serverBlock.SetText(serverText)
+		blocks = append(blocks, serverBlock)
+	}
+
+	return blocks, nil
+}
```
droot, line 203: Would be nice if we can decouple the UI from a core package like mcp.

I wonder if can represent MCP server status as another structure and render the UI blocks from that structure outside of this package.

I would also like if we can introduce special command `mcp` in our repl to display the mcp server status etc.

cmd/mcp_client.go
```diff
@@ -0,0 +1,252 @@
...
+
+// StartMCPServer starts the MCP server with the given configuration
+func StartMCPServer(ctx context.Context, opt Options) error {
```
droot, line 206: since we have mcp server and client both in this file, may be keeping the file name `mcp.go` is probably better

droot (commented):
Thank you @tuannvm . This is such a well done PR.

I took a quick look and shared early comments.

I will take deeper look tomorrow, but this is already looking solid, so shouldn't take much iteration.

I will also think of some compelling scenario that we can share in screencast when we announce this feature.

README.md
tuannvm: I don’t see why not. In fact, dealing with JSON in Cursor is quite a pain. Why don’t we support both formats? Converting between JSON and YAML isn’t a big deal.

pkg/agent/systemprompt_template_default.txt
tuannvm: The way I see it, this would help chain multiple steps together. For example:

- "Use sequential thinking to analyze the current naming convention from the list of namespaces."

The LLM would execute the steps in this order:

- Call the `kubectl` tool and run `kubectl get namespace`.
- Automatically call the `sequential thinking` tool to perform the analysis.

in single prompt

cmd/mcp_client.go
tuannvm: I think keeping them separate would help with cross-collaboration and provide a clear separation between the MCP client and MCP server, which might otherwise be confusing for new users.

README.md
tuannvm: However, we need to be mindful of our users. If they are coming from Cursor or Windsurf, they might prefer to use the same configuration format as their IDE. Perhaps we could offer a tool to convert configurations to YAML?

droot:
@tuannvm Will do a final pass tomorrow. Thank you so much!

cmd/mcp_client.go
droot: Sounds good.

README.md
droot: Generally we would like to avoid two ways to doing the same things to keep UX simple.
Keeping the same config format is also important (kubectl config, tools config and mcp config).

pkg/tools/mcp_tool.go
```diff
@@ -0,0 +1,301 @@
...
+	// No automatic discovery in init() to ensure proper control flow
+}
```
droot, line 60: in the spirit of explicit control, let's use instances instead of globals. Might be better to get rid of this init block.

pkg/tools/mcp_tool.go
```diff
@@ -0,0 +1,301 @@
...
+	"fmt"
+	"strconv"
+	"strings"
+
+	"github.com/GoogleCloudPlatform/kubectl-ai/gollm"
+	"github.com/GoogleCloudPlatform/kubectl-ai/pkg/mcp"
+	"k8s.io/klog/v2"
+)
+
+var mcpManager *mcp.Manager
```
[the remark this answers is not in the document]
droot: if any other part of the stack is causing issues (say tool registry), we can change that.

pkg/tools/mcp_tool.go
```diff
@@ -0,0 +1,301 @@
...
+	}
+	return value // Keep original if conversion fails
+}
+
+// =============================================================================
+// MCP Integration Functions
+// =============================================================================
+
+// registerToolsFromConnectedServers discovers tools from all connected MCP servers and registers them
+func registerToolsFromConnectedServers(ctx context.Context) error {
```
droot, line 221: this can be a method on the mcpManager instance.

pkg/tools/mcp_tool.go
```diff
@@ -0,0 +1,301 @@
...
+}
+
+// RefreshMCPTools forces a refresh of MCP tools (for manual refresh scenarios)
+func RefreshMCPTools(ctx context.Context) error {
+	if mcpManager == nil {
+		return fmt.Errorf("MCP manager not initialized")
+	}
+
+	return registerToolsFromConnectedServers(ctx)
+}
```
droot, line 301: All these can become methods on the manager instance and we can hold the instance as dependency in main.go during bootstrap.

pkg/mcp/utils.go
```diff
@@ -0,0 +1,256 @@
...
+	"context"
+	"fmt"
+	"math"
+	"os"
+	"os/exec"
+	"path/filepath"
+	"strings"
+	"time"
+
+	mcp "github.com/mark3labs/mcp-go/mcp"
```
droot, line 27: wondering if this package was useful ?

pkg/mcp/status.go
```diff
@@ -0,0 +1,190 @@
...
+}
+
+// MCPManager defines the interface needed by GetServerStatus to interact with MCP
+type MCPManager interface {
+	ListClients() []*Client
+	ListAvailableTools(ctx context.Context) (map[string][]Tool, error)
+}
+
+// GetServerStatus returns status information for all MCP servers
+func GetServerStatus(ctx context.Context, mcpClientEnabled bool, mcpManager MCPManager) (*MCPStatus, error) {
```
droot, line 57: All these methods should be anchored on the mcpManager instance.

droot (changes requested):
This is looking very good. I have two major concerns that we need to address:

1. Using YAML as the config format (this will be important because we will eventually merge or make it easier to specify configs for kubectl-ai, mcp and custom tools).
2. Move away from global (single pattern).

README.md
```diff
@@ -201,6 +201,48 @@ A custom tool definition for `helm` could look like the following example:
...
+
+The system automatically:
+- Converts parameter names (snake_case → camelCase)
+- Handles type conversion (strings → numbers/booleans when appropriate)
+- Provides fallback behavior for unknown servers
+
+No additional setup required - just use the `--mcp-client` flag and the AI will have access to all configured MCP tools.
+
+📖 **For detailed configuration options, troubleshooting, and advanced features, see the [MCP Documentation](pkg/mcp/README.md).**
```
janetkuo, line 244: There's also https://github.com/GoogleCloudPlatform/kubectl-ai/blob/main/docs/mcp.md, linked from https://github.com/GoogleCloudPlatform/kubectl-ai?tab=readme-ov-file#mcp-server
Need to disambiguate them instead of just calling both MCP (kubectl-ai as a MCP server vs. client)

pkg/mcp/README.md
```diff
@@ -0,0 +1,168 @@
...
+
+## Overview
+
+The MCP client allows `kubectl-ai` to connect to MCP servers, discover available tools, and execute them. This enables integration with various services and systems that expose their functionality through the MCP protocol.
+
+## Features
+
+- Connect to multiple MCP servers simultaneously
+- Automatic discovery of available tools from connected servers
+- Execute tools on remote MCP servers with parameter conversion
```
droot, line 13: is there an example of remote mcp server that we can try out ?

running command locally is already supported through custom tools, so good differentiation will be able to use it with a remote mcp server.

pkg/mcp/client.go
```diff
@@ -0,0 +1,365 @@
...
+
+	// Build proper environment slice by merging process environment with custom env
+	finalEnv := mergeEnvironmentVariables(os.Environ(), c.Env)
+
+	return expandedCmd, finalEnv, nil
+}
+
+// createMCPClient creates the underlying MCP client
+func (c *Client) createMCPClient(command string, env []string) error {
+	client, err := mcpclient.NewStdioMCPClient(command, env, c.Args...)
```
droot, line 109: Aah...doesn't look like we are supporting remote mcp client... right ?

droot (commented):
some clarifications.

pkg/mcp/client.go
tommynguyen-vungle: You’re right. I overlooked this. I’ll add a `TODO` for adding SSE support.

tommynguyen-vungle: You’re right. I overlooked this. I’ll add a `TODO` for adding SSE support.

pkg/mcp/README.md
```diff
@@ -0,0 +1,168 @@
...
+- The configuration file has strict permissions (0600) by default
+- Be cautious when adding environment variables with sensitive information
+
+## Troubleshooting
+
+### Common Issues
+
+**No MCP tools available:**
+- Ensure you're using the `--mcp-client` flag
+- Check that `~/.config/kubectl-ai/mcp.yaml` exists and is valid
```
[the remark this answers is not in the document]
tuannvm: Adjusted. I think it’s good to explicitly call out the dependencies, even if they are created by default.

droot (approved):

pkg/mcp/client.go
tuannvm: yes, stdio only for now. Please see comment above.

pkg/mcp/config.go
```diff
@@ -0,0 +1,283 @@
...
+	"k8s.io/klog/v2"
+	"sigs.k8s.io/yaml"
+)
+
+// ServerConfig represents the configuration for an MCP server
+type ServerConfig struct {
+	// Name is a friendly name for this MCP server
+	Name string `yaml:"name"`
+	// Command is the command to execute for stdio-based MCP servers
+	Command string `yaml:"command"`
```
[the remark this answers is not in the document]
tuannvm: Thanks @droot, acked. Yes, let’s go ahead and merge it first, please.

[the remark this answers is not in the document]
droot: Ok, getting this in. Thanks.

tuannvm:
> @tuannvm I was about to get it merged and then found out presubmits (go-build and go-vet) are still failing. Can you pl. take a look. Thanks.

@droot please check again. Should be okay now. There is additional IsInteractive method so I just add a placeholder version for mcp_tools.go for now

pkg/mcp/config.go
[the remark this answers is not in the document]
nisranjan: > @droot I have a working version of the streamable HTTP/SSE MCP server, but I’ll need some time to fine-tune it. Shall we merge the stdio-based approach first? I’ve added a TODO here:
> 
> https://github.com/GoogleCloudPlatform/kubectl-ai/blob/78701ee745d614950b04240a7d63cc3090d1cc62/pkg/mcp/client.go#L43

@droot @tuannvm
So sounds like we want a stream-able HTTP/SSE MCP Server. This is the open item on this PR?

Issue #254, opened 2025-05-20
Warning logged when missing default tool config


hakman:
When starting `kubectl-ai`, without defining any custom tools, a warning is logged:
```
W0520 18:28:59.913875   97471 main.go:321] Failed to load or register custom tools (path:
  ["{CONFIG}/kubectl-ai/tools.yaml" "{HOME}/.config/kubectl-ai/tools.yaml"]):
  failed to describe config file /Users/hakman/.config/kubectl-ai/tools.yaml:
  stat /Users/hakman/.config/kubectl-ai/tools.yaml: no such file or directory
````

Default location(s) for custom tools configuration should not trigger a warning when not found.
For more details, see https://github.com/GoogleCloudPlatform/kubectl-ai/pull/246#discussion_r2096902561.

Relevant code sections:
https://github.com/GoogleCloudPlatform/kubectl-ai/blob/b8e4c66e27f3235e0999c1d908eb98bae448b700/cmd/main.go#L299-L323
https://github.com/GoogleCloudPlatform/kubectl-ai/blob/b8e4c66e27f3235e0999c1d908eb98bae448b700/pkg/tools/tools.go#L191-L194

PR #271, opened 2025-05-23
Fixed tool config handling

droot:
Summary:

1. Fixed a bug in resolving the paths for custom-tool-configs. `filepath.Join([]string{"", "a", "b"})` will result in "a/b" instead of "/a/b".
2. If we fail to read the user specified custom tool configs (non-defaults), we will return error immediately. fixes https://github.com/GoogleCloudPlatform/kubectl-ai/issues/254

### DIFF
--- a/cmd/main.go
+++ b/cmd/main.go
@@ -18,6 +18,7 @@ import (
 	"bufio"
 	"bytes"
 	"context"
+	"errors"
 	"flag"
 	"fmt"
 	"io"
@@ -25,6 +26,7 @@ import (
 	"os"
 	"os/signal"
 	"path/filepath"
+	"slices"
 	"strings"
 	"syscall"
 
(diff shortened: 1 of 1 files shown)


droot:
@hakman I was trying out custom tools and discovered a few issues. Would be great if you can take a look since you are most familiar with this code.

hakman:
Sure, will take a look later today.

janetkuo (approved):
Looks good!

cmd/main.go
```diff
@@ -130,6 +132,16 @@ func (u *UserInterface) Type() string {
...
 
+var defaultToolConfigPaths = []string{
+	filepath.Join("{CONFIG}", "kubectl-ai", "tools.yaml"),
+	filepath.Join("{HOME}", ".config", "kubectl-ai", "tools.yaml"),
+}
+
+var defaultConfigPaths = []string{
+	filepath.Join("{CONFIG}", "kubectl-ai", "config.yaml"),
+	filepath.Join("{HOME}", ".config", "kubectl-ai", "config.yaml"),
+}
```
hakman, line 143: I think we can simplify this by having a single variable named `defaultLookupPaths` and then just appending the config or tools file name.

PR #272, opened 2025-05-23
cleanup: deduplicate common instruction logic in `systemprompt_template_default.txt`

yankay:
Removed the conditional block (`if .EnableToolUseShim`) and its associated instructions, simplifying the template and ensuring the prompt is more concise.


droot (approved):
Thank you :)

PR #273, opened 2025-05-23
FIX: kubectl streaming and interaction bashtool #122 #232

### LINKED ISSUE #122: FR: please support streaming of logs

### LINKED ISSUE #232: Bug | kubectl -w loops infinitely
While testing, I came up to a scenario where the model executes 

  Running: kubectl get pod parserp-pod -w

Creating an infinite loop, Here is a forced scenario simulating it

```sh
❯ ./kubectl-ai --llm-provider=grok --model=grok-3-beta "create a pod from zvdy/parserp:latest and then watch for updates with -w"

  I'm going to create a pod using the image  zvdy/parserp:latest  and set up a
  watch to monitor updates as requested.                                      
                                                                              
  Let me analyze the possible approaches to solve this task:                  
                                                                              
  1. Direct Pod Creation with Watch: Use  kubectl run  to create a pod and    
  immediately follow it with a  kubectl get pod -w  to watch for updates. This
  is straightforward but might require two separate commands.                 
  2. Pod Creation via YAML: Create a YAML manifest for the pod and apply it,  
  then use  kubectl get pod -w  to watch. This gives more control over pod    
  configuration but is more complex for a simple task.                        
  3. Use Kubectl Run with Watch Option: Check if  kubectl run  can be combined
  with a watch option directly. This would be ideal but might not be supported
  natively in one command.                                                    
  4. Scripted Solution: Write a small bash script to create the pod and then  
  watch it. This adds complexity and is unnecessary for a single pod.         
  5. Deployment Instead of Pod: Create a deployment instead of a pod for      
  better management and watch it. This might be overkill for a simple pod     
  request.                                                                    
  6. Immediate Watch After Creation: Create the pod with  kubectl run  and in 
  the same command pipeline or immediately after, watch it with  kubectl get  
  pod -w . This ensures the watch starts right after creation.                
  7. Check for Existing Pod: Before creating, check if a pod with the same    
  image or name exists to avoid duplication, then create and watch. This adds 
  an extra step but prevents errors.                                          
                                                                              
  I believe the best approach is a combination of Direct Pod Creation with    
  Watch (Option 1) and Immediate Watch After Creation (Option 6). This ensures
  the pod is created and watched in a streamlined manner without unnecessary  
  complexity. I'll use  kubectl run  to create the pod quickly and then watch 
  it with  kubectl get pod -w .                                               
                                                                              
  Let me proceed with creating the pod and setting up the watch.              


  Running: kubectl run parserp-pod --image=zvdy/parserp:latest --restart=Never

  Do you want to proceed ?
  1) Yes
  2) Yes, and don't ask me again
  3) No
  Enter your choice (number): 1

  The pod  parserp-pod  has been successfully created with the image          
  zvdy/parserp:latest . Now, I'll set up a watch to monitor updates to the pod
  as requested.                                                               


  Running: kubectl get pod parserp-pod -w

```

Only option here would be to `Control+C`

## Solutions

- Edit the system-prompt https://github.com/GoogleCloudPlatform/kubectl-ai/blob/269eeecd05b14ca6627cb3c97cfb48b8d2179ea7/pkg/agent/systemprompt_template_default.txt#L45-L52
- Add a timeout at some point (although this possibly breaks other scenarios)
- Add this edge case in documentation

zvdy:
# Fix streaming command handling with timeout

## Summary
Implements a simple timeout-based approach for handling streaming kubectl commands. It addresses issues with commands like `kubectl logs -f` and `kubectl get -w` by automatically terminating them after 7 seconds, while still allowing users to see the output in real-time.

## Changes
- Added detection for various streaming kubectl commands:
  - `kubectl logs -f`
  - `kubectl get -w`
  - `kubectl attach`
- Explicitly blocked interactive commands:
  - `kubectl exec -it`
  - `kubectl port-forward`
- Implemented a 7-second timeout for all streaming commands
- Added proper stream type identification for different operations
- Simplified the command execution logic by removing complex input handling

## Testing
```
>>> ^C
❯ ./kubectl-ai --llm-provider=openai --model=claude-3-5-sonnet-20241022 "create a pod in the default namespace that outputs to stdout a random 6char string every 4 secs"

  I'll help you create a pod that generates and outputs random 6-character    
  strings every 4 seconds. I'll use a shell script with  tr  and  head        
  commands to generate random strings, running in a container.                
                                                                              
  Let me create this pod using  kubectl :                                     


  Running: cat << 'EOF' | kubectl apply -f -
apiVersion: v1
kind: Pod
metadata:
  name: random-string-generator
  namespace: default
spec:
  containers:
  - name: random-generator
    image: busybox:latest
    command: ["/bin/sh", "-c"]
    args:
      - while true; do
          head -c 32 /dev/urandom | tr -dc 'a-zA-Z0-9' | head -c 6;
          echo;
          sleep 4;
        done
EOF

  Do you want to proceed ?
  1) Yes
  2) Yes, and don't ask me again
  3) No
  Enter your choice (number): 1

  Great! I've created a pod with the following characteristics:               
                                                                              
  1. Name: random-string-generator                                            
  2. Uses busybox image (lightweight and contains necessary tools)            
  3. Runs a shell script that:                                                
    • Reads random data from /dev/urandom                                     
    • Filters it to only alphanumeric characters                              
    • Takes 6 characters                                                      
    • Outputs the string                                                      
    • Sleeps for 4 seconds                                                    
    • Repeats indefinitely                                                    
                                                                              
                                                                              
  Let's verify that the pod is running and check its logs to see the output:  


  Running: kubectl get pod random-string-generator


  Let's see the output:                                                       


  Running: kubectl logs -f random-string-generator --tail=5
laXnwl
2z8vyP
TO0dk

Timeout reached after 7 seconds


  Perfect! 🎉 The pod is running successfully and generating random 6-        
  character                                                                   
  strings every 4 seconds. You can continue to monitor the output using:      
                                                                              
    kubectl logs -f random-string-generator                                   
                                                                              
  The pod will continue running and generating random strings until it's      
  deleted. Would you like me to show you more output or help you with anything
  else?                                                                       

```

```
❯ ./kubectl-ai --llm-provider=openai --model=claude-3-5-sonnet-20241022 "can we exec into random-string-generator pod in default?"

  I'll help you exec into the random-string-generator pod. Let me first verify
  the pod exists and is running.                                              


  Running: kubectl get pod random-string-generator -n default


  Great! The pod exists and is running. Let me execute into the pod. I'll use 
  the default container since there appears to be only one container (READY   
  1/1).                                                                       


  Running: kubectl exec -it random-string-generator -n default -- sh


  I apologize for the confusion. It seems the interactive terminal session is 
  not supported through this interface. However, I can help you run specific  
  commands within the pod if you'd like. For example, I could:                
                                                                              
  1. List the contents of a directory                                         
  2. Check environment variables                                              
  3. Run a specific command                                                   
  4. View running processes                                                   
  5. Check the application logs                                               
                                                                              
  Would you like me to run any specific command inside the pod? Please let me 
  know what you'd like to do within the container, and I'll help you execute  
  that command non-interactively.                                             

```

## Related Issues
Closes #122 - Need to print outputs of commands that run in a loop
Closes #232 - Ability to gracefully exit from streaming commands

## Notes
- Tried using Control+C but the repl and signals we're giving me implementation issues,
- 7 is an arbitrary number, i had the idea to make it interactive to, so prompt user to add a timeout, but perhaps someone wants to run kubectl-ai automatically or dokcerized as mentioned in some feature requests
- I also had an idea to add a global flag to it, so the timeout being 7 and allowing it to be customize it

I'm open to modifications maybe a mix of the two?
![image](https://github.com/user-attachments/assets/65b2d607-d9d1-4ea6-a297-d6259a3c566f)

### DIFF
--- a/pkg/agent/conversation.go
+++ b/pkg/agent/conversation.go
@@ -171,7 +171,6 @@ func (a *Conversation) RunOneRound(ctx context.Context, query string) error {
 		}
 
 		// Process each part of the response
-		// only applicable is not using tooluse shim
 		var functionCalls []gollm.FunctionCall
 
 		var agentTextBlock *ui.AgentTextBlock
@@ -210,10 +209,12 @@ func (a *Conversation) RunOneRound(ctx context.Context, query string) error {
 					agentTextBlock.AppendText(text)
 				}
 
-				// Check if it's a function call
-				if calls, ok := part.AsFunctionCalls(); ok && len(calls) > 0 {
-					log.Info("function calls", "calls", calls)
-					functionCalls = append(functionCalls, calls...)
+				// Only process function calls if not using tool-use shim
+				if !a.EnableToolUseShim {
+					if calls, ok := part.AsFunctionCalls(); ok && len(calls) > 0 {
+						log.Info("function calls", "calls", calls)
+						functionCalls = append(functionCalls, calls...)
+					}
 				}
 			}
 		}
@@ -222,112 +223,114 @@ func (a *Conversation) RunOneRound(ctx context.Context, query string) error {
 			agentTextBlock.SetStreaming(false)
 		}
 
-		// TODO(droot): Run all function calls in parallel
-		// (may have to specify in the prompt to make these function calls independent)
-		for _, call := range functionCalls {
-			toolCall, err := a.Tools.ParseToolInvocation(ctx, call.Name, call.Arguments)
-			if err != nil {
-				return fmt.Errorf("building tool call: %w", err)
-			}
-
-			// Check if the command is interactive using the tool's implementation
-			isInteractive, errMsg := toolCall.GetTool().IsInteractive(call.Arguments)
-			klog.Infof("isInteractive: %t, errMsg: %s", isInteractive, errMsg)
-
-			// If interactive, handle based on whether we're using tool-use shim
-			if isInteractive {
-				if a.EnableToolUseShim {
-					// For models without tool-use support, append as text response
-					if agentTextBlock == nil {
-						agentTextBlock = ui.NewAgentTextBlock()
-						a.doc.AddBlock(agentTextBlock)
-					}
-					agentTextBlock.AppendText(fmt.Sprintf("  %s\n", errMsg))
-				} else {
-					// For models with tool-use support, use proper FunctionCallResult
-					errorBlock := ui.NewErrorBlock().SetText(fmt.Sprintf("  %s\n", errMsg))
-					a.doc.AddBlock(errorBlock)
-
-					currChatContent = append(currChatContent, gollm.FunctionCallResult{
-						ID:     call.ID,
-						Name:   call.Name,
-						Result: map[string]any{"error": errMsg},
-					})
+		// Only process function calls if not using tool-use shim
+		if !a.EnableToolUseShim {
+			// TODO(droot): Run all function calls in parallel
+			for _, call := range functionCalls {
+				toolCall, err := a.Tools.ParseToolInvocation(ctx, call.Name, call.Arguments)
+				if err != nil {
+					return fmt.Errorf("building tool call: %w", err)
 				}
-				continue
-			}
 
-			// Only show "Running" message and proceed with execution for non-interactive commands
-			s := toolCall.PrettyPrint()
-			a.doc.AddBlock(ui.NewFunctionCallRequestBlock().SetText(fmt.Sprintf("  Running: %s\n", s)))
+				// Check if the command is interactive using the tool's implementation
+				isInteractive, errMsg := toolCall.GetTool().IsInteractive(call.Arguments)
+				klog.Infof("isInteractive: %t, errMsg: %s", isInteractive, errMsg)
 
-			var result any
-			var invokeErr error
+				// If interactive, handle based on whether we're using tool-use shim
+				if isInteractive {
+					if a.EnableToolUseShim {
+						// For models without tool-use support, append as text response
+						if agentTextBlock == nil {
+							agentTextBlock = ui.NewAgentTextBlock()
+							a.doc.AddBlock(agentTextBlock)
+						}
+						agentTextBlock.AppendText(fmt.Sprintf("  %s\n", errMsg))
+					} else {
+						// For models with tool-use support, use proper FunctionCallResult
+						errorBlock := ui.NewErrorBlock().SetText(fmt.Sprintf("  %s\n", errMsg))
+						a.doc.AddBlock(errorBlock)
 
-			// Execute the command
-			result, invokeErr = toolCall.InvokeTool(ctx, tools.InvokeToolOptions{
-				Kubeconfig: a.Kubeconfig,
-				WorkDir:    a.workDir,
-			})
-			if invokeErr != nil {
-				return fmt.Errorf("executing action: %w", invokeErr)
-			}
+						currChatContent = append(currChatContent, gollm.FunctionCallResult{
+							ID:     call.ID,
+							Name:   call.Name,
+							Result: map[string]any{"error": errMsg},
+						})
+					}
+					continue
+				}
 
-			// Convert result to map for the LLM
-			resultMap, err := tools.ToolResultToMap(result)
-			if err != nil {
-				return err
-			}
+				// Only show "Running" message and proceed with execution for non-interactive commands
+				s := toolCall.PrettyPrint()
+				a.doc.AddBlock(ui.NewFunctionCallRequestBlock().SetText(fmt.Sprintf("  Running: %s\n", s)))
 
-			// Handle timeout message using UI blocks
-			if execResult, ok := result.(*tools.ExecResult); ok && execResult.StreamType == "timeout" {
-				a.doc.AddBlock(ui.NewAgentTextBlock().WithText("\nTimeout reached after 7 seconds\n"))
-			}
+				var result any
+				var invokeErr error
 
-			// Add the tool call result to maintain conversation flow
-			currChatContent = append(currChatContent, gollm.FunctionCallResult{
-				ID:     call.ID,
-				Name:   call.Name,
-				Result: resultMap,
-			})
+				// Execute the command
+				result, invokeErr = toolCall.InvokeTool(ctx, tools.InvokeToolOptions{
+					Kubeconfig: a.Kubeconfig,
+					WorkDir:    a.workDir,
+				})
+				if invokeErr != nil {
+					return fmt.Errorf("executing action: %w", invokeErr)
+				}
 
-			// Only ask for confirmation if SkipPermissions is false and command modifies a resource
-			if !a.SkipPermissions && call.Arguments["modifies_resource"] != "no" {
-				confirmationPrompt := `  Do you want to proceed ?
+				// Convert result to map for the LLM
+				resultMap, err := tools.ToolResultToMap(result)
+				if err != nil {
+					return err
+				}
+
+				// Handle timeout message using UI blocks
+				if execResult, ok := result.(*tools.ExecResult); ok && execResult.StreamType == "timeout" {
+					a.doc.AddBlock(ui.NewAgentTextBlock().WithText("\nTimeout reached after 7 seconds\n"))
+				}
+
+				// Add the tool call result to maintain conversation flow
+				currChatContent = append(currChatContent, gollm.FunctionCallResult{
+					ID:     call.ID,
+					Name:   call.Name,
+					Result: resultMap,
+				})
+
+				// Only ask for confirmation if SkipPermissions is false and command modifies a resource
+				if !a.SkipPermissions && call.Arguments["modifies_resource"] != "no" {
+					confirmationPrompt := `  Do you want to proceed ?
   1) Yes
   2) Yes, and don't ask me again
   3) No`
 
-				optionsBlock := ui.NewInputOptionBlock().SetPrompt(confirmationPrompt)
-				optionsBlock.SetOptions([]string{"1", "2", "3"})
-				a.doc.AddBlock(optionsBlock)
+					optionsBlock := ui.NewInputOptionBlock().SetPrompt(confirmationPrompt)
+					optionsBlock.SetOptions([]string{"1", "2", "3"})
+					a.doc.AddBlock(optionsBlock)
 
-				selectedChoice, err := optionsBlock.Observable().Wait()
-				if err != nil {
-					if err == io.EOF {
-						// Use hit control-D, or was piping and we reached the end of stdin.
-						// Not a "big" problem
-						return nil
+					selectedChoice, err := optionsBlock.Observable().Wait()
+					if err != nil {
+						if err == io.EOF {
+							// Use hit control-D, or was piping and we reached the end of stdin.
+							// Not a "big" problem
+							return nil
+						}
+						return fmt.Errorf("reading input: %w", err)
 					}
-					return fmt.Errorf("reading input: %w", err)
-				}
 
-				switch selectedChoice {
-				case "1":
-					// Proceed with the operation
-				case "2":
-					a.SkipPermissions = true
-				case "3":
-					a.doc.AddBlock(ui.NewAgentTextBlock().WithText("Operation was skipped."))
-					observation := fmt.Sprintf("User didn't approve running %q.\n", call.Name)
-					currChatContent = append(currChatContent, observation)
-					continue
-				default:
-					// This case should technically not be reachable due to AskForConfirmation loop
-					err := fmt.Errorf("invalid confirmation choice: %q", selectedChoice)
-					log.Error(err, "Invalid choice received from AskForConfirmation")
-					a.doc.AddBlock(ui.NewErrorBlock().SetText("Invalid choice received. Cancelling operation."))
-					return err
+					switch selectedChoice {
+					case "1":
+						// Proceed with the operation
+					case "2":
+						a.SkipPermissions = true
+					case "3":
+						a.doc.AddBlock(ui.NewAgentTextBlock().WithText("Operation was skipped."))
+						observation := fmt.Sprintf("User didn't approve running %q.\n", call.Name)
+						currChatContent = append(currChatContent, observation)
+						continue
+					default:
+						// This case should technically not be reachable due to AskForConfirmation loop
+						err := fmt.Errorf("invalid confirmation choice: %q", selectedChoice)
+						log.Error(err, "Invalid choice received from AskForConfirmation")
+						a.doc.AddBlock(ui.NewErrorBlock().SetText("Invalid choice received. Cancelling operation."))
+						return err
+					}
 				}
 			}
 		}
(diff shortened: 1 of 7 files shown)


pkg/tools/bash_tool.go
```diff
@@ -127,13 +130,97 @@ func (t *BashTool) Run(ctx context.Context, args map[string]any) (any, error) {
...
+	isWatch := strings.Contains(command, "kubectl get") && strings.Contains(command, "-w")
+	isLogs := strings.Contains(command, "kubectl logs") && strings.Contains(command, "-f")
+	isExec := strings.Contains(command, "kubectl exec") && strings.Contains(command, "-it")
+	isAttach := strings.Contains(command, "kubectl attach")
+	isPortForward := strings.Contains(command, "kubectl port-forward")
+
+	// Block interactive commands
+	if isExec || isPortForward {
+		return &ExecResult{Error: "interactive mode not supported for kubectl, please use non-interactive commands"}, nil
+	}
```
droot, line 151: I have noticed that we will ask for user's permission to run a command and then later apply this filtering. Which is confusing and annoying. So I would suggest:
If we can refactor this filtering in a separate function, then we can call that function even before we ask for approval from users.

Also: Its okay to just introduce the filtering in a common function here, but actual change to avoid asking for approval for interactive/blocking command in a separate follow up PR.

droot (changes requested):
This is already a big improvements. I have minor comments.

Thanks for taking this up.

pkg/tools/bash_tool.go
```diff
@@ -127,13 +130,97 @@ func (t *BashTool) Run(ctx context.Context, args map[string]any) (any, error) {
...
+
```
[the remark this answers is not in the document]
zvdy: Did not think about this, this improved a lot the llm behavior d927fc5d3a2aeb055f49c9c7bcdb73e6da2a58fa

pkg/tools/bash_tool.go
zvdy: Forgotten to test this out, it makes sense though, still, this is almost always avoided with the latest commits:

```
"Can you exec into hello-pod?" 


  I'll help you check and exec into the hello-pod. Let me first verify if this
  pod exists and its status.                                                  


  Running: kubectl get pod hello-pod


  Great! I can see that hello-pod exists and is running. Let me execute a     
  simple command to verify we can connect to it. Since we should avoid        
  interactive sessions, I'll use a non-interactive exec command to check the  
  shell environment:                                                          


  Running: kubectl exec hello-pod -- /bin/sh -c "echo 'Successfully connected to hello-pod'"


  ✅ I've successfully connected to hello-pod! Since you haven't specified    
  what                                                                        
  command you'd like to run inside the pod, I've just verified that we can    
  execute commands in it.                                                     
                                                                              
  If you'd like to run specific commands inside the pod, please let me know   
  what commands you'd like to execute. For example, you could ask to:         
                                                                              
  • Check the filesystem                                                      
  • View running processes                                                    
  • Check environment variables                                               
  • Execute specific applications                                             
                                                                              
  What would you like to do inside the pod?                                   

```

pkg/agent/conversation.go
```diff
@@ -230,9 +230,64 @@ func (a *Conversation) RunOneRound(ctx context.Context, query string) error {
 				return fmt.Errorf("building tool call: %w", err)
 			}
 
+			// Check if this is a kubectl or bash command that might be interactive
+			isInteractive := false
+			var errMsg string
+			if call.Name == "kubectl" || call.Name == "bash" {
```
droot, line 236: I wonder if we can introduce a method in the `Tool` interface called `IsInteractive(args) bool` to delegate this logic to individual tools to keep the core agentic loop tool agnostic.

pkg/agent/conversation.go
```diff
@@ -230,9 +230,64 @@ func (a *Conversation) RunOneRound(ctx context.Context, query string) error {
...
+				}
+			}
+
+			// If interactive, show error and skip to next iteration
+			if isInteractive {
+				// Use ErrorBlock for interactive command errors
+				errorBlock := ui.NewErrorBlock().SetText(fmt.Sprintf("  %s\n", errMsg))
+				a.doc.AddBlock(errorBlock)
+
+				currChatContent = append(currChatContent, gollm.FunctionCallResult{
```
droot, line 248: So models that doesn't support tool-use yet, we need to condition this logic on `EnableToolUseShim` (deleted lines in line 274....).

So if `EnableToolUseShim` --> Just append as text response else proper FunctionCallResult (like you have today).


I think `EnableToolUseShim` can be pushed at the gollm layer, but that is some for other day. But I am also hoping that we will probably drop this soon because most models will support tool-use by default.

pkg/agent/conversation.go
```diff
@@ -230,9 +230,64 @@ func (a *Conversation) RunOneRound(ctx context.Context, query string) error {
...
+				return err
+			}
+
+			// Handle timeout message using UI blocks
+			if execResult, ok := result.(*tools.ExecResult); ok && execResult.StreamType == "timeout" {
+				a.doc.AddBlock(ui.NewAgentTextBlock().WithText("\nTimeout reached after 7 seconds\n"))
+			}
+
+			// Add the tool call result to maintain conversation flow
+			currChatContent = append(currChatContent, gollm.FunctionCallResult{
```
droot, line 284: I think this need to be conditioned to `EnableToolUseShim` like we had earlier.

pkg/tools/bash_tool.go
```diff
@@ -127,13 +130,139 @@ func (t *BashTool) Run(ctx context.Context, args map[string]any) (any, error) {
...
-	Stderr   string `json:"stderr,omitempty"`
-	ExitCode int    `json:"exit_code,omitempty"`
+	Error      string `json:"error,omitempty"`
+	Stdout     string `json:"stdout,omitempty"`
+	Stderr     string `json:"stderr,omitempty"`
+	ExitCode   int    `json:"exit_code,omitempty"`
+	StreamType string `json:"stream_type,omitempty"`
+}
+
+func IsInteractiveCommand(command string) (bool, string) {
```
droot, line 140: I suggested if we make this method part of Tool definition. Will make it cleaner and eventually play well with the custom tool support as well.

Or eval asking another LLM to tell us if given command is interactive or not.

pkg/tools/kubectl_tool.go
```diff
@@ -35,7 +34,17 @@ func (t *Kubectl) Name() string {
 }
 
 func (t *Kubectl) Description() string {
-	return "Executes a kubectl command against the user's Kubernetes cluster. Use this tool only when you need to query or modify the state of the user's Kubernetes cluster."
+	return `Executes a kubectl command against the user's Kubernetes cluster. Use this tool only when you need to query or modify the state of the user's Kubernetes cluster.
+
+IMPORTANT: Interactive commands are not supported in this environment. This includes:
+- kubectl exec with -it flag (use non-interactive exec instead)
+- kubectl edit (use kubectl get -o yaml and kubectl apply instead)
```
droot, line 41: I have noticed when asking LLM for an alternative to `edit`, they come up with `patch` a lot. Would suggest including patch also in the hint`.

droot (changes requested):
I like where we are going with this, so thank you :) Took a quick look and have a few minor suggestions. let me know what you think.

pkg/agent/conversation.go
```diff
@@ -210,10 +209,12 @@ func (a *Conversation) RunOneRound(ctx context.Context, query string) error {
 					agentTextBlock.AppendText(text)
 				}
 
-				// Check if it's a function call
-				if calls, ok := part.AsFunctionCalls(); ok && len(calls) > 0 {
-					log.Info("function calls", "calls", calls)
-					functionCalls = append(functionCalls, calls...)
+				// Only process function calls if not using tool-use shim
+				if !a.EnableToolUseShim {
```
droot, line 213: So we don't need to condition this. If toolUseShim is enabled, we convert the textual LLM response to a ShimCandidate response that has function calls etc. (line 165 above). Think of it as: client side implementation of tool-use since server side (API) doesnt' support tool-use.

pkg/agent/conversation.go
```diff
@@ -222,108 +223,114 @@ func (a *Conversation) RunOneRound(ctx context.Context, query string) error {
...
-			}
-
-			// Check if this is a kubectl or bash command that might be interactive
-			isInteractive := false
-			var errMsg string
-			if call.Name == "kubectl" || call.Name == "bash" {
-				if command, ok := call.Arguments["command"].(string); ok {
-					isInteractive, errMsg = tools.IsInteractiveCommand(command)
+		// Only process function calls if not using tool-use shim
+		if !a.EnableToolUseShim {
```
droot, line 227: We don't need to condition. We have functionCalls for both (tool-use or non-tool-use API) at this point.

pkg/agent/conversation.go
```diff
@@ -222,108 +223,114 @@ func (a *Conversation) RunOneRound(ctx context.Context, query string) error {
...
+							agentTextBlock = ui.NewAgentTextBlock()
+							a.doc.AddBlock(agentTextBlock)
+						}
+						agentTextBlock.AppendText(fmt.Sprintf("  %s\n", errMsg))
+					} else {
+						// For models with tool-use support, use proper FunctionCallResult
+						errorBlock := ui.NewErrorBlock().SetText(fmt.Sprintf("  %s\n", errMsg))
+						a.doc.AddBlock(errorBlock)
+
+						currChatContent = append(currChatContent, gollm.FunctionCallResult{
```
droot, line 253: Only this part needs to be conditioned to `tool-use-shim`

pkg/agent/conversation.go
```diff
@@ -222,108 +223,114 @@ func (a *Conversation) RunOneRound(ctx context.Context, query string) error {
...
-				})
-				continue
-			}
+				// Check if the command is interactive using the tool's implementation
+				isInteractive, errMsg := toolCall.GetTool().IsInteractive(call.Arguments)
+				klog.Infof("isInteractive: %t, errMsg: %s", isInteractive, errMsg)
+
+				// If interactive, handle based on whether we're using tool-use shim
+				if isInteractive {
+					if a.EnableToolUseShim {
```
droot, line 241: Rendering the text block is not conditional to tool-use or not.

droot (commented):
I think I messed up in conveying the changes related to tool-use-shim.

Here is a quick explainer:

some models don't support tool-use in the API (for ex. gemma3).

tool-use-shim sort of uses different prompt to steer the model to give us tool-call in a special format (we call it ReAct style).

In the agentic loop, we convert the response to normalized Candidate so that rest of the code can treat function call like normally. However when we add the response back, our code doesn't abstract properly and we have if/else in the response we are returning.

pkg/agent/conversation.go
```diff
@@ -171,7 +171,6 @@ func (a *Conversation) RunOneRound(ctx context.Context, query string) error {
 		}
```
droot, line 171: Another idea in response to most recent changes (making tool-use-shim-aware), we can probably have fnResultToShimFunctionResult at the end and ignore tool-id etc. if `EnableToolUseShim`, that way atleast process part of the response will be agnostic to the tool-use-shim or not.

zvdy:
# Tests with gemini

```
❯ ./kubectl-ai --model=gemini-2.5-flash-preview-04-17 "Can you stream the logs of the random-string-generator?"


  Okay, I can help you with that. I will try to stream the logs of the        
  resource named  random-string-generator . I will assume it's a pod for now, 
  as                                                                          
  that's the most common resource type for streaming logs.                    


  Running: kubectl logs -f random-string-generator
ArK
PEW
xXL
YZD
Zqs
eoV
kwY
oyF
tyZ
aha
DvH
ioq
FHs
efZ
Elx
nDI
RCp
rJQ
Zoc


  Timeout reached after 7 seconds                                             


```

```
❯ ./kubectl-ai --model=gemini-2.5-flash-preview-04-17 "Can you exec into the random-string-generator pod?"


  Interactive  kubectl exec  is not supported in this environment. If you want
  to execute a specific command non-interactively inside the  random-string-  
  generator  pod, please provide the command you would like to run.           

```

pkg/agent/conversation.go
```diff
@@ -223,15 +222,44 @@ func (a *Conversation) RunOneRound(ctx context.Context, query string) error {
...
 		// TODO(droot): Run all function calls in parallel
-		// (may have to specify in the prompt to make these function calls independent)
 		for _, call := range functionCalls {
 			toolCall, err := a.Tools.ParseToolInvocation(ctx, call.Name, call.Arguments)
 			if err != nil {
 				return fmt.Errorf("building tool call: %w", err)
 			}
 
+			// Check if the command is interactive using the tool's implementation
+			isInteractive, errMsg := toolCall.GetTool().IsInteractive(call.Arguments)
```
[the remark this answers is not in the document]
zvdy: the interface was a string, but agree, error makes sense

zvdy:
No problem! :-)

Let me know if adding a flag `--stream-timeout` or `--command-timeout` would also be a good thing to implement

Although `7s` is quite generous for most workloads

Issue #280, opened 2025-05-26
Fix: make `config.yaml` Optional


yankay:
After the https://github.com/GoogleCloudPlatform/kubectl-ai/pull/271  change,
If there is no `config.yaml` file in the environment, kubectl-ai will fail to run.

```
root@gpu3090:~/oss/kubectl-ai# task run
task: [run] go run ./cmd
failed to load config file: could not load config from "/root/.config/kubectl-ai/config.yaml": open /root/.config/kubectl-ai/config.yaml: no such file or directory
exit status 1
task: Failed to run task "run": exit status 1
```

This behavior is unexpected, so we should revert this part of the code to its previous state,
making it ​​ignore missing config files​​ (i.e., treat them as ​​optional​​).

----
kubectl-ai version: main branch

PR #281, opened 2025-05-26
Fix: make config.yaml Optional 

yankay:
Fix: #280

After the https://github.com/GoogleCloudPlatform/kubectl-ai/pull/271  change,
If there is no `config.yaml` file in the environment, kubectl-ai will fail to run.

```
root@gpu3090:~/oss/kubectl-ai# task run
task: [run] go run ./cmd
failed to load config file: could not load config from "/root/.config/kubectl-ai/config.yaml": open /root/.config/kubectl-ai/config.yaml: no such file or directory
exit status 1
task: Failed to run task "run": exit status 1
```

This behavior is unexpected, so we should revert this part of the code to its previous state,
making it ​​ignore missing config files​​ (i.e., treat them as ​​optional​​).

----
kubectl-ai version: main branch

### DIFF
--- a/cmd/main.go
+++ b/cmd/main.go
@@ -198,9 +198,8 @@ func (o *Options) LoadConfigurationFile() error {
 		configPath = filepath.Clean(pathWithPlaceholdersExpanded)
 		configBytes, err := os.ReadFile(configPath)
 		if err != nil {
-			if os.IsNotExist(err) && !slices.Contains(defaultConfigPaths, configPath) {
-				// user specified config file does not exist
-				return fmt.Errorf("could not load config from %q: %w", configPath, err)
+			if os.IsNotExist(err) {
+				// ignore missing config files, they are optional
 			} else {
 				fmt.Fprintf(os.Stderr, "warning: could not load defaults from %q: %v\n", configPath, err)
 			}


droot:
Ouch, my bad, I introduced this bug. I do wanted it to fail if user specified any config that don't exist. That will require another change, but lets get this in first.

Thank you.

droot:
@yankay you have been contributing to k8s related projects for a while, was wondering if you have any feedback on the project ?

droot:
> I'm currently learning about kubectl-ai, which is an awesome project. Therefore, I'd like to get more involved. I'm not sure if this project has a mechanism similar to the Kubernetes Community membership, as described here: https://github.com/kubernetes/community/blob/master/community-membership.md#reviewer. This would allow new contributors to participate and eventually become Reviewers and Approvers.

Thanks for trying out the project. We have not been able to spend much time on membership aspects of the project yet, but we definitely will with time. We are seeing very healthy engagement from contributors.

> * The Kubernetes documentation (https://kubernetes.io/docs) and the Issues, PRs, and KEP discussions in the Kubernetes GitHub repositories (https://github.com/kubernetes/*) have accumulated a wealth of knowledge and experience. If this knowledge could be leveraged through technologies like RAG (Retrieval-Augmented Generation) or WebSearch, it might enhance the capabilities of kubectl-ai.

Good idea.  RAG comes with it's own set of challenges but I think asking model to use `WebSearch` (in the API side itself) would be something interesting to try out and compare. Let me know if you are interested in taking this up. We can create an issue and geek out on it.

PR #283, opened 2025-05-26
feat(cli): Adding a Description to UserInterface in the CLI

yankay:
Add a description for UserInterface In the CLI, "User interface mode to use. Supported values: terminal, html."

Because currently there is no documentation for the usage of `html`, it can only be seen by referring to the code.


droot (approved):
html interface is still work in progress, but doesn't hurt to make it discoverable.

PR #286, opened 2025-05-27
fix of main command was getting lost

### LINKED ISSUE #274: bug: kubectil-ai invokes incomplete commands (without the main command)
Sometimes `kubectl-ai` will invoke tools without the main command (as prefix), for example:

```sh

# without kubectl prefix
Running: get pods -n web

```

Even though we specify in the tool description that `command` is a complete command including the `kubectl` prefix, some models tend to be bad at instruction following. It also suggests we have a gap in our instructions for tool use.

I recently tried custom tool and discovered top tier foundation models were running into this problem, so I suspect we are missing something basic here.

In this [video](https://www.youtube.com/watch?v=LFQpuZb_r6s), you can see user encounters it repeatedly with qwen-coder model with ollama provider.

A few ideas for exploration:

1. Tweak the main prompt
2. Tweak the tool-user or function-calling instructions
3. Additional checks in the code to ensure not to invoke without the main command (this could be an issue with composite commands (pipe syntax etc.)

selimacerbas:
fixes #274 

I tested on llama3.2:latest and llama3.2:1b.

I tried tweaking system prompt, kubectl description and bash description as well but non of them was helping LLM to use necessary main command. 

For that reason added sanitisation logic after tool call to add the main command in front of arguments.

Any better ideas? @droot @hakman 

Thanks in advance!

### DIFF
--- a/pkg/agent/conversation.go
+++ b/pkg/agent/conversation.go
@@ -233,6 +233,9 @@ func (a *Conversation) RunOneRound(ctx context.Context, query string) error {
 				return fmt.Errorf("building tool call: %w", err)
 			}
 
+			// Sanitize arguments (e.g. enforce kubectl prefix, etc.)
+			sanitizeToolArguments(toolCall.GetTool().Name(), call.Arguments)
+
 			// Check if the command is interactive using the tool's implementation
 			isInteractive, err := toolCall.GetTool().IsInteractive(call.Arguments)
 			klog.Infof("isInteractive: %t, err: %v, CallArguments: %+v", isInteractive, err, call.Arguments)
@@ -442,6 +445,25 @@ func extractJSON(s string) (string, bool) {
 	data := s[first+len(jsonBlockMarker) : last]
 
 	return data, true
+}
+
+// Applies tool-specific sanity checks and corrections incase some
+// commands are not visible. This has been mostly experienced in local small LLM models.
+func sanitizeToolArguments(toolName string, args map[string]any) {
+	switch toolName {
+	case "kubectl":
+		// auto-prepend missing "kubectl " prefix
+		if cmd, ok := args["command"].(string); ok && !strings.HasPrefix(cmd, "kubectl ") {
+			klog.Warningf("kubectl missing prefix; auto-prepending")
+			args["command"] = "kubectl " + cmd
+		}
+
+	// case "gdrive":
+	//     // future logic for gdrive, etc.
+
+	default:
+		// no-op
+	}
 }
 
 // parseReActResponse parses the LLM response into a ReActResponse struct


pkg/agent/conversation.go
```diff
@@ -444,6 +447,25 @@ func extractJSON(s string) (string, bool) {
 	return data, true
 }
 
+// Applies tool-specific sanity checks and corrections incase some
+// commands are not visible. This has been mostly experienced in local small LLM models.
+func sanitizeToolArguments(toolName string, args map[string]any) {
```
droot, line 452: I think the main issue is that, sometimes valid commands may not start with `kubectl` .. think command that uses heredoc syntax or composite commands.

echo "...." |kubectl apply -f 

honestly, I am questioning if we should really try to address this. It is actually a problem of the model not support tool-calling well and eventually will get better.

None of the foundation models run in to this, so I am questioning if it makes sense to fight this.

hakman: I also 100% agree with this, let's keep things simple.
I will send a proposal to make it easier to at least understand what tool is ran, but have to experiment a little there.

droot: Good feedback. Building on what you guys have already said:

We will see this pattern again and again where frontier models will be more accurate than smaller or old models. Our natural instinct will to try to patch these edge cases like this. This may work in the short term, but will be irrelevant in the long term. AI researchers and practitioners refers to this as [the bitter lessons](http://www.incompleteideas.net/IncIdeas/BitterLesson.html). So, frontier model performance will be the leading indicator and it is probably safe to assume that open models will eventually follow.
If this is the reality, then how should we go about it. A few things that will help:

- Comprehensive evals (`k8s-bench`) that shows different model's performance for k8s tasks and we also track it over time. And when the end-users are making a choice on what model to use, they can look at the `k8s-bench` and can at least rule out some models to avoid this frustration or claiming `kubectl-ai` or AI in general to be useless. And if we have to make decide on patching these edge cases, evals can inform that decision as well.
- for those rare cases, where we have to patch these edge cases, we can reduce the pain for us maintainers by good engineering. In this specific case, good engineering will be for ex. have a good parser for the commands we are invoking. We have had instances where we had to filter a few commands (for ex. detecting interactive or streaming commands). I was looking more into it and found there is a robust bash parsing library [mvdan/sh](https://github.com/mvdan/sh) that can be used to inspect the tool-call better and then patch it more robustly.
- Can we make smaller model to perform better for our (k8s and infra) domain ?. I think this where a partnership with AI researchers will help. There are techniques such as distilling knowledge from frontier models as synthetic data and then fine-tuning the smaller models with that synthetic data. And evals will help here to verify the results.

@selimacerbas It may appear that the PR is going no where, but this has already helped the project in a big way IMO in helping us build better understanding. When you build stuff at the edge (frontier), you will see gaps before others see it and these gaps are what makes the difference and sort of guide you how to move the field forward (progress).

This reminds me of an essay from Paul Graham [Live in the future, and then build what's missing](https://paulgraham.com/startupideas.html).

PR #301, opened 2025-06-02
feat: Update google.golang.org/genai to v1.8.0

droot:
I've updated the `google.golang.org/genai` dependency from v1.4.0 to v1.8.0 in the `gollm` module.

This change also required updating the `go.sum` file in the `gollm` module and running `go mod tidy` in both the `gollm` module and the root module to ensure all dependencies are consistent.

The root module now correctly lists `google.golang.org/genai v1.8.0` as an indirect dependency.


mikebz (approved):

mikebz:
tried to speed up this review but it looks like I don't have write access.

PR #309, opened 2025-06-03
fix/feat | Static Kubectl Command Filter Implementation

### LINKED ISSUE #276: bug: kubectl-ai asks for approval even when command doesn't modify a resource
`kubectl-ai` uses LLM to determine if a given command modifies a resource in kubernetes or not. We have seen foundation models such as Gemini, openai etc. gets this right, but local models tend to hallucinate more in this aspect.

For example, in this [video](https://www.youtube.com/watch?v=LFQpuZb_r6s), there are instances where no resource is being modified but the agent will still ask for approval.

I think we can definitely improve here. I am wondering if we can maintain a static filter for kubectl commands to determine if it modifies a resource of not or explore some other direction.

/cc @zvdy @hakman @tuannvm @selimacerbas @Vinay-Khanagavi

zvdy:
## Summary
This PR implements a static filter system to determine if kubectl commands modify Kubernetes resources, replacing the previous LLM-based approach. This addresses the inconsistencies and potential hallucinations noted in issue #276.

### Pattern
The implementation uses a combined whitelist/blacklist pattern:
- **Whitelist** of read-only operations (e.g., `get`, `describe`, `logs`)
- **Blacklist** of resource-modifying operations (e.g., `apply`, `create`, `delete`)
- **Special case handling** for complex scenarios like:
  - Commands with `--dry-run` flags (never modify resources)
  - Multi-part commands like `config set-context`
  - Kubectl plugins with known behaviors
  - Output redirection and watch operations

### Classification Results
Commands are classified into three categories:
- `"yes"` - Command definitely modifies resources
- `"no"` - Command definitely doesn't modify resources  
- `"unknown"` - Command's effect can't be determined with certainty

### Fallback Behavior
When the static filter returns `"unknown"`:
- The system defaults to treating the command as potentially modifying resources
- User confirmation will be requested (unless `SkipPermissions` is enabled)
- This conservative approach ensures safety for commands that can't be definitively classified, _I rather fail to be over-secure than modify a unwanted resource from someone without confirmation_

## Testing
I also included unit tests; 
> _I plan to add more if I get the free time and eventually add a step in the `ci-presubmit` that triggers all of the tests before merge_
- Standard kubectl commands (both modifying and non-modifying)
- Commands with dry-run flags
- Multi-word commands
- Edge cases with environment variables and path variations
- Various command formatting variations

## Impact
This change improves kubectl-ai's accuracy in determining when to ask for user confirmation:
- Works with both shim-enabled and shim-disabled modes
- No changes needed to conversation.go as it already respected the `modifies_resource` parameter

## Closes
Closes #276 - "Improve detection of kubectl commands that modify resources"
# Log example

`I0603 19:36:54.794082  173609 conversation.go:241] isInteractive: false, err: <nil>, CallArguments: map[command:kubectl delete pod -n dev-zvdy -l app=nginx-ingress modifies_resource:yes]
`

### DIFF
--- a/pkg/tools/kubectl_filter.go
+++ b/pkg/tools/kubectl_filter.go
@@ -172,28 +172,3 @@ func hasDryRunFlag(command string) bool {
 	}
 	return false
 }
-
-// containsAny checks if the command contains any of the given strings
-// For flag detection, it properly checks word boundaries to avoid false positives
-func containsAny(command string, substrings []string) bool {
-	// Split the command into space-separated tokens to properly detect flags
-	tokens := strings.Fields(command)
-
-	for _, substr := range substrings {
-		// Simple substring check for non-flag strings
-		if !strings.HasPrefix(substr, "--") {
-			if strings.Contains(command, substr) {
-				return true
-			}
-			continue
-		}
-
-		// For flags, we need to ensure proper flag matching
-		for _, token := range tokens {
-			if token == substr || strings.HasPrefix(token, substr+"=") {
-				return true
-			}
-		}
-	}
-	return false
-}
--- a/pkg/tools/kubectl_filter_test.go
+++ b/pkg/tools/kubectl_filter_test.go
@@ -176,28 +176,6 @@ func TestKubectlAnalyzerComponents(t *testing.T) {
 		}
 	})
 
-	t.Run("containsAny detection", func(t *testing.T) {
-		tests := []struct {
-			command    string
-			substrings []string
-			expected   bool
-		}{
-			{"kubectl get pods -w", []string{"-w", "--watch"}, true},
-			{"kubectl get pods --watch", []string{"-w", "--watch"}, true},
-			{"kubectl get pods", []string{"-w", "--watch"}, false},
-			{"kubectl get pods -o wide", []string{"-o", "--output"}, true},
-			{"kubectl get pods --output=json", []string{"--output"}, true},
-			{"kubectl get pods -output=json", []string{"--output"}, false}, // Wrong flag format
-		}
-
-		for _, tt := range tests {
-			result := containsAny(tt.command, tt.substrings)
-			if result != tt.expected {
-				t.Errorf("containsAny(%q, %v) = %v, want %v", tt.command, tt.substrings, result, tt.expected)
-			}
-		}
-	})
-
 	t.Run("command parsing", func(t *testing.T) {
 		tests := []struct {
 			command     string


zvdy:
> [!NOTE]
> Perhaps it's worth changing the system prompt too, and rely strictly on the kubectl_filter, this could save some tokens

This involves changing `Action` struct;

https://github.com/GoogleCloudPlatform/kubectl-ai/blob/0869096f453323e16361fb44c5fc56ebf86003ec/pkg/agent/conversation.go#L430-L435

zvdy:
> _Will have a look at https://github.com/mvdan/sh, and refactor if this simplifies it, still, we're simply parsing kubectl commands here, logic will remain similar_

pkg/tools/bash_tool.go
```diff
@@ -103,6 +103,16 @@ func (t *BashTool) Run(ctx context.Context, args map[string]any) (any, error) {
 	workDir := ctx.Value(WorkDirKey).(string)
 	command := args["command"].(string)
 
+	// Always set modifies_resource value for kubectl commands
```
droot: `modifies_resource` field is used in the caller to seek approval from the user before invoking the tool. Updating the value here wouldn't help with the overall flow unless I am missing something here.

pkg/tools/bash_tool.go
```diff
@@ -103,6 +103,16 @@ func (t *BashTool) Run(ctx context.Context, args map[string]any) (any, error) {
...
+	if strings.Contains(command, "kubectl") {
+		args["modifies_resource"] = KubectlModifiesResource(command)
+		klog.Infof("Set modifies_resource=%s for command: %s", args["modifies_resource"], command)
+	} else {
+		// For non-kubectl commands, default to unknown
```
droot: This is more of a self-note. I think we haven't thought much about what are the semantics for this field for non-kubectl commands. Each tool use does read/write stuff in the environments that they are interacting with (file-system, k8s or say cloud), so far we have only captured k8s, but the notion need to be extended to non-k8s as well but currently it is confusing.

pkg/tools/kubectl_filter.go
```diff
@@ -0,0 +1,241 @@
...
+import (
+	"strings"
+)
+
+// KubectlModifiesResource determines if a kubectl command modifies resources.
+// Returns:
+// - "yes" if the command definitely modifies resources
+// - "no" if the command definitely doesn't modify resources
+// - "unknown" if we can't determine for sure
+func KubectlModifiesResource(command string) string {
```
droot: I am thinking if it makes sense to add a method to our tool interface (similar to how we handled interactive commands):

```Go

type Tool interface {

...

ModifiesResource(args ...) 

...

}
```

Then in the caller (conversation.go), we can check in a tool agnostic way how we do it for interactive commands.

pkg/tools/kubectl_filter.go
```diff
@@ -0,0 +1,241 @@
...
+
+	// List of prefixes/keywords that indicate resource modification
+	modifyingOperations := []string{
+		"create", "apply", "edit", "delete", "patch", "replace",
+		"scale", "autoscale", "expose", "rollout", "run", "set",
+		"label", "annotate", "taint", "drain", "cordon", "uncordon",
+		"certificate approve", "certificate deny", "debug", "attach",
+		"cp", "exec", "port-forward", "auth reconcile",
+		"alpha events trigger",
+	}
```
droot: I am curious how did you construct these lists ? (is there a way to ensure that this list is exhaustive) ?

pkg/tools/kubectl_filter.go
```diff
@@ -0,0 +1,241 @@
...
+
+	// Find the position of kubectl in case there are prefixes (like KUBECONFIG=... kubectl)
+	kubectlPos := -1
+	for i, part := range cmdParts {
+		if strings.HasSuffix(part, "kubectl") {
+			kubectlPos = i
+			break
+		}
+	}
```
droot, line 82: This is making me think if we should use solid shell parser ?

pkg/tools/kubectl_filter.go
droot: building on what I said above,

We also need to remove the `modifies_resource` argument from the function-calling/tool-use.

I think this will simplify the logic even further.

pkg/tools/kubectl_filter.go
zvdy: yeah, mentioned this on the pull previously, for interactive stuff i didn't mind as it was quite simple enough but having a look at  https://github.com/mvdan/sh makes sense to use it, trusted and simplifies this process a lot, not only for kubectl but general parsing for future or clients tools in general, will add it on the kubectl_filter

pkg/tools/kubectl_filter.go
zvdy: my source was: https://kubernetes.io/docs/reference/generated/kubectl/kubectl-commands

I haven't deep-dived, but covers all/most of these, and if something is missing. or newly implemented it would be as easy as adding it to the `[]string`

pkg/tools/bash_tool.go
zvdy: realised this too, this aligns with one of your comments below, but the plan was:

> Perhaps it's worth changing the system prompt too, and rely strictly on the kubectl_filter, this could save some tokens

so not relying in 
https://github.com/GoogleCloudPlatform/kubectl-ai/blob/eadf4485337210cd1641158452f641e82b807fa1/pkg/agent/systemprompt_template_default.txt#L22

even, skipping this and saving some tokens by doing it with the `tools`

pkg/tools/kubectl_filter.go
zvdy: Agree on this, removing it from tools + implementing the interface makes it easier to maintain. follows DRY principle

zvdy:
## Recent Commit Changes

1. **Robust Shell Parsing**  
   Uses `mvdan/sh` to accurately parse complex kubectl commands, handling quotes, pipes, variables, and multi-command statements.

2. **Standardized Danger Checks via Interface**  
   All tools now implement a unified `CheckModifiesResource(args)` interface, returning `"yes"`, `"no"`, or `"unknown"` to indicate if a command modifies cluster resources. This standardizes permission checks across all tool types.

3. **Comprehensive Command Classification**  
   Updated and expanded lists of read-only and modifying kubectl operations (see kubectl_filter.go), including edge cases and plugins, for more accurate detection.

4. **Permission Flow Fixes**  
   Before running any command, the agent uses the new interface to check if it modifies resources and prompts the user for confirmation unless `SkipPermissions` is set.

5. **Deterministic Results & Fallback Logic**  
   The filter aims to avoid `"unknown"` except for ambiguous cases (e.g., `kubectl exec`). If the tool’s check returns `"unknown"`, the system can optionally fall back to the LLM’s own assessment for a second opinion, ensuring a safe default.

6. **Extensive Unit Testing**  
   Added in-depth tests for edge cases, multi-command scenarios, plugins, and ambiguous commands to ensure reliability.

---

Let me know if any of the tests do not make sense, or if responses should be revised based on your criteria.

> cc: @droot

zvdy:
> [!NOTE]
removing the llm makes no sense _(to me after testing and digging into it)_, for commands like `exec` or simply the mcp,

It will need to be persisted and backwards compatible, and used as fallback if the deterministic approach is not sufficient, we cant filter  all unix and common OSS tool commands, but, we can guardrail most of the kubectl ones as that is the intention of the interface cli tool.

for instance;

`{"Exec command", "kubectl exec nginx -- ls", "unknown"},`

why unknown? perhaps instead of ls we execute any other command `rm -rf whatever` that modifies filesystem, should be asked, so better to have the LLM handle it.

So although I initially wanted to remove that load from the LLM and perhaps save on some tokens, I think handling it this way is more beneficial, most cases will pass through `KubectlModifiesResource` and get marked as yes/no

pkg/agent/conversation.go
```diff
@@ -267,7 +267,21 @@ func (a *Conversation) RunOneRound(ctx context.Context, query string) error {
...
+			// If our code detection returned "unknown", fall back to the LLM's assessment if available
+			if modifiesResourceStr == "unknown" {
+				// For tool-use shim mode, check the Action.ModifiesResource field
+				if a.EnableToolUseShim && call.Name == "bash" || call.Name == "kubectl" {
+					if llmModifies, ok := call.Arguments["modifies_resource"].(string); ok {
+						klog.Infof("Code detection returned 'unknown', falling back to LLM assessment: %s", llmModifies)
+						modifiesResourceStr = llmModifies
+					}
+				}
+			}
```
droot, line 279: we can simply say:

```Go

if modifiesResourceStr == "unknown" {
      if llmModifies, ok := call.Arguments["modifies_resource"].(string); ok {
			klog.Infof("Code detection returned 'unknown', falling back to LLM assessment: %s", llmModifies)
			modifiesResourceStr = llmModifies
	}
}

```
because `modifies_resource` is applicable whether or not tool-use-shim is enabled or not.

pkg/agent/conversation.go
```diff
@@ -267,7 +267,21 @@ func (a *Conversation) RunOneRound(ctx context.Context, query string) error {
...
+
+			if !a.SkipPermissions && modifiesResourceStr == "yes" {
```
droot: We need to be more conservative here, unless the `modifiesResourceStr` is `no`, we should assume the command modifies the resource and requires approval (like we had earlier) (basically account for "unknown" and "yes").

```
!a.SkipPermissions && call.Arguments["modifies_resource"] != "no"
```

pkg/tools/bash_tool.go
```diff
@@ -84,15 +84,6 @@ func (t *BashTool) FunctionDefinition() *gollm.FunctionDefinition {
...
 				},
-				"modifies_resource": {
-					Type: gollm.TypeString,
-					Description: `Whether the command modifies a kubernetes resource.
-Possible values:
-- "yes" if the command modifies a resource
-- "no" if the command does not modify a resource
-- "unknown" if the command's effect on the resource is unknown
-`,
-				},
```
droot: sometimes LLM will use bashtool to run kubectl command (composite command). Since we fallback on LLM to determine if resource is modified or not, this is still needed.

pkg/tools/custom_tool.go
```diff
@@ -72,15 +72,6 @@ func (t *CustomTool) FunctionDefinition() *gollm.FunctionDefinition {
...
 				},
-				"modifies_resource": {
-					Type: gollm.TypeString,
-					Description: `Whether the command modifies a kubernetes resource.
-Possible values:
-- "yes" if the command modifies a resource
-- "no" if the command does not modify a resource
-- "unknown" if the command's effect on the resource is unknown
-`,
-				},
```
droot: I wonder if we have to change anything in the custom tool implementation to account for this change.

I am thinking if we should make this generic and remove kubernetes from here. because if I using `gcloud` or some other tool, approval flow is still useful.

pkg/tools/kubectl_tool.go
```diff
@@ -82,15 +82,6 @@ kubectl apply -f pod.yaml
...
 				},
-				"modifies_resource": {
-					Type: gollm.TypeString,
-					Description: `Whether the command modifies a kubernetes resource.
-Possible values:
-- "yes" if the command modifies a resource
-- "no" if the command does not modify a resource
-- "unknown" if the command's effect on the resource is unknown
-`,
-				},
```
droot, line 93: Since we fallback to LLM, we still need this.

droot (commented):
Some early comments. Need to take deeper look at the filter code.

pkg/tools/custom_tool.go
zvdy: Did not think about cloud resource modifications, makes sense to keep it imo, llm fallback method can look into those request and determine if yes/no

pkg/agent/conversation.go
zvdy: Makes sense to over ask, makes it safer although it might mark false positives for non kubectl tools, still, better to have false positives than modified resources without confirmation

pkg/agent/conversation.go
```diff
@@ -267,6 +267,17 @@ func (a *Conversation) RunOneRound(ctx context.Context, query string) error {
...
+
+			// If our code detection returned "unknown", fall back to the LLM's assessment if available
+			if modifiesResourceStr == "unknown" {
+				if llmModifies, ok := call.Arguments["modifies_resource"].(string); ok {
+					klog.Infof("Code detection returned 'unknown', falling back to LLM assessment: %s", llmModifies)
+					modifiesResourceStr = llmModifies
+				}
+			}
+
 			if !a.SkipPermissions && call.Arguments["modifies_resource"] != "no" {
```
droot: s/call.Arguments["modifies_resource"]/modifiesResourceStr

pkg/tools/kubectl_filter.go
```diff
@@ -0,0 +1,604 @@
...
+)
+
+// analyzeWithShellParser uses the mvdan/sh parser to analyze kubectl commands
+// Returns the resource modification status and whether parsing was successful
+func analyzeWithShellParser(command string) (string, bool) {
+	// Parse the command using mvdan/sh
+	parser := syntax.NewParser()
+	file, err := parser.Parse(strings.NewReader(command), "")
+	if err != nil {
+		klog.V(4).Infof("Failed to parse command with mvdan/sh: %v", err)
```
droot: why `V(4)` ? we should log this as error.

pkg/tools/kubectl_filter.go
```diff
@@ -0,0 +1,604 @@
...
+// - "no" if the command definitely doesn't modify resources
+// - "unknown" if we can't determine for sure
+func KubectlModifiesResource(command string) string {
+	// Try to use the shell parser to analyze the command
+	result, parsed := analyzeWithShellParser(command)
+	if parsed {
+		return result
+	}
+
+	// Fall back to the traditional string-based method if shell parsing fails
```
droot: Isn't it safe to assume the command is invalid if parsing fails ?

droot (changes requested):
Need to take deeper look at the kubectl filter code.

pkg/tools/kubectl_filter.go
```diff
@@ -0,0 +1,604 @@
...
+
+		// For flags, we need to ensure proper flag matching
+		for _, token := range tokens {
+			if token == substr || strings.HasPrefix(token, substr+"=") {
+				return true
+			}
+		}
+	}
+	return false
+}
```
droot, line 174: I was curious and spent some time in cursor today. The writeVerbs needs to be made exhaustive but the core logic seems to be simpler and smaller.

```Go
package main

import (
	"fmt"
	"strings"

	"mvdan.cc/sh/v3/syntax"
)

func main() {
	// Test cases for write operation detection
	testCases := []string{
		// Simple write operations
		`kubectl apply -f config.yaml`,
		`kubectl create deployment nginx --image=nginx`,
		`kubectl delete pod mypod`,
		`kubectl replace -f config.yaml`,
		`kubectl edit deployment nginx`,

		// Write operations with dry-run
		`kubectl apply -f config.yaml --dry-run=client`,
		`kubectl create deployment nginx --image=nginx --dry-run`,
		`kubectl delete pod mypod --dry-run=server`,

		// Composite commands
		`cat config.yaml | kubectl apply -f -`,
		`kubectl get pods --dry-run=client && kubectl delete pod mypod`,
		`kubectl apply -f config.yaml || echo "Failed"`,

		// Read-only operations
		`kubectl get pods`,
		`kubectl describe pod mypod`,
		`kubectl logs mypod`,

		// Exec commands
		`kubectl exec -it mypod -- ls -l`,
		`kubectl exec -it mypod -- cat /etc/passwd`,
		`kubectl exec -it mypod -- rm -rf /`,
		`kubectl exec -it mypod -- bash`,
		`kubectl exec -it mypod -- curl http://api`,
		`kubectl exec -it mypod -- wget http://file`,
		`kubectl exec -it mypod -- echo "test" > /tmp/test`,
	}

	for _, cmd := range testCases {
		isWrite, reason := isKubectlWriteOperation(cmd)
		fmt.Printf("\nCommand: %s\n", cmd)
		fmt.Printf("Is Write Operation: %v\n", isWrite)
		if reason != "" {
			fmt.Printf("Reason: %s\n", reason)
		}
	}
}

// List of known kubectl write operations
var writeVerbs = map[string]bool{
	"apply":     true,
	"create":    true,
	"delete":    true,
	"replace":   true,
	"edit":      true,
	"patch":     true,
	"scale":     true,
	"set":       true,
	"rollout":   true,
	"autoscale": true,
	"exec":      true,
}

func isKubectlWriteOperation(cmd string) (bool, string) {
	parser := syntax.NewParser()
	file, err := parser.Parse(strings.NewReader(cmd), "")
	if err != nil {
		return false, fmt.Sprintf("Error parsing command: %v", err)
	}

	var writeCommands []struct {
		verb      string
		hasDryRun bool
	}

	syntax.Walk(file, func(node syntax.Node) bool {
		call, ok := node.(*syntax.CallExpr)
		if !ok || len(call.Args) == 0 {
			return true
		}

		// Check if the command is 'kubectl'
		if call.Args[0].Lit() == "kubectl" {
			hasDryRun := false
			var writeVerb string

			// Check for dry-run flag
			for _, arg := range call.Args[1:] {
				lit := arg.Lit()
				if strings.HasPrefix(lit, "--dry-run") {
					hasDryRun = true
					break
				}
			}

			// Check for write verbs
			for _, arg := range call.Args[1:] {
				lit := arg.Lit()
				if lit == "" || strings.HasPrefix(lit, "-") {
					continue // skip flags and empty
				}
				if writeVerbs[lit] {
					writeVerb = lit
					break
				}
			}

			if writeVerb != "" {
				writeCommands = append(writeCommands, struct {
					verb      string
					hasDryRun bool
				}{writeVerb, hasDryRun})
			}
		}
		return true
	})

	if len(writeCommands) == 0 {
		return false, ""
	}

	// Check if any write command doesn't have dry-run
	for _, cmd := range writeCommands {
		if !cmd.hasDryRun {
			return true, fmt.Sprintf("Command contains write verb '%s' without --dry-run flag", cmd.verb)
		}
	}

	// All write commands have dry-run
	return false, fmt.Sprintf("All write commands (%v) have --dry-run flag",
		func() []string {
			var verbs []string
			for _, cmd := range writeCommands {
				verbs = append(verbs, cmd.verb)
			}
			return verbs
		}())
}

```

pkg/tools/kubectl_filter.go
zvdy: We fallback to string parsing if it fails, we can remove that and rely strictly and only on `mvdan`

that's why its not an err

pkg/tools/custom_tool.go
```diff
@@ -70,7 +71,7 @@ func (t *CustomTool) FunctionDefinition() *gollm.FunctionDefinition {
 			Properties: map[string]*gollm.Schema{
 				"command": {
 					Type:        gollm.TypeString,
-					Description: t.config.CommandDesc,
+					Description: `The bash command to execute.`,
```
droot, line 74: Is this change intentional ?
Custom tool can have their own description that can help LLM make better decision about the tool.

droot: infact, we should tweak the description of `modifies_resource` below to make it kubernetes agnostic.

pkg/tools/kubectl_filter.go
```diff
@@ -0,0 +1,251 @@
...
+
+	if len(args) == 0 {
+		klog.Warning("analyzeCall: no arguments extracted from call")
+		return commandResult{unknownFound: true}
+	}
+
+	// Find kubectl command
+	kubectlPos := -1
+	for i, arg := range args {
+		if arg == "kubectl" || arg == "/usr/local/bin/kubectl" {
```
droot: for checking `/usr/local/bin/kubectl`, we should make it os agnostic (won't work on windows). There are utilities in either os or path, that helps with (basically last part of the command needs to be `kubectl`).

pkg/tools/kubectl_filter.go
```diff
@@ -0,0 +1,251 @@
...
+			kubectlPos = i
+			break
+		}
+		if strings.Contains(arg, "kubectl") {
+			klog.V(2).Infof("analyzeCall: ambiguous kubectl arg: %q", arg)
+			return commandResult{unknownFound: true}
+		}
+	}
+
+	if kubectlPos == -1 || kubectlPos >= len(args)-1 {
```
droot: Since we are analyzing a single callExpr, I am wondering why not just check the args[0] to be "kubectl" (or has last element to be kubectl).

pkg/tools/kubectl_filter.go
```diff
@@ -0,0 +1,251 @@
...
+		return commandResult{unknownFound: true}
+	}
+
+	// Get the verb (first non-flag argument after kubectl)
+	verbPos := kubectlPos + 1
+	for verbPos < len(args) && strings.HasPrefix(args[verbPos], "-") {
+		verbPos++
+	}
+
+	if verbPos >= len(args) {
```
droot, line 142: I think this index based logic is very brittle, I think we can avoid it since iterating over a slice is all we need.

pkg/tools/kubectl_filter.go
```diff
@@ -0,0 +1,251 @@
...
+				klog.V(2).Infof("analyzeCall: special case write for verb=%q subcmd=%q", verb, args[verbPos+1])
+				return commandResult{writeFound: true}
+			}
+			klog.V(2).Infof("analyzeCall: special case read for verb=%q subcmd=%q", verb, args[verbPos+1])
+			return commandResult{readFound: true}
+		}
+	}
+
+	// Check plugins
+	if result, ok := knownPlugins[verb]; ok {
```
droot: I think special casing krew plugins is a slippery slope and we should avoid it. It is totally fine to defer it to the user to make the decision and not handle this special case (or leave it to LLM to make a call on it).

pkg/tools/kubectl_filter.go
```diff
@@ -0,0 +1,251 @@
...
+		klog.V(2).Infof("analyzeCall: known plugin read for verb=%q", verb)
+		return commandResult{readFound: true}
+	}
+
+	// Check standard operations
+	if writeOps[verb] && !hasDryRun {
+		klog.V(2).Infof("analyzeCall: write op for verb=%q", verb)
+		return commandResult{writeFound: true}
+	}
+	if readOnlyOps[verb] || (writeOps[verb] && hasDryRun) {
```
droot, line 157: Why we need to track readOnly commands, all we care about is writeOnly.

pkg/tools/kubectl_filter.go
```diff
@@ -0,0 +1,251 @@
...
+		"label": true, "annotate": true, "taint": true, "drain": true,
+		"cordon": true, "uncordon": true, "certificate": true,
+		"debug": true, "attach": true, "cp": true,
+	}
+
+	specialCases = map[string]map[string]string{
+		"auth": {
+			"can-i": "no", "whoami": "no", "reconcile": "yes",
+		},
+		"certificate": {
```
droot: this can simply be added to writeOps

pkg/tools/kubectl_filter.go
```diff
@@ -0,0 +1,251 @@
...
+		"create": true, "apply": true, "edit": true, "delete": true,
+		"patch": true, "replace": true, "scale": true, "autoscale": true,
+		"expose": true, "rollout": true, "run": true, "set": true,
+		"label": true, "annotate": true, "taint": true, "drain": true,
+		"cordon": true, "uncordon": true, "certificate": true,
+		"debug": true, "attach": true, "cp": true,
+	}
+
+	specialCases = map[string]map[string]string{
+		"auth": {
```
droot: we can simply track the writeOnly (in this case "reconcile").

pkg/tools/kubectl_filter.go
```diff
@@ -0,0 +1,251 @@
...
+var (
+	readOnlyOps = map[string]bool{
+		"get": true, "describe": true, "explain": true, "top": true,
+		"logs": true, "api-resources": true, "api-versions": true,
+		"version": true, "config": true, "cluster-info": true,
+		"wait": true, "auth": true, "diff": true, "kustomize": true,
+		"help": true, "options": true, "plugin": true, "proxy": true,
+		"completion": true, "convert": true, "alpha": true, "events": true,
+		"port-forward": true,
+	}
```
droot, line 34: I don't think we need to track readonly

pkg/tools/kubectl_filter.go
```diff
@@ -0,0 +1,251 @@
...
+	return commandResult{unknownFound: true}
+}
+
+func normalizeCommand(command string) string {
+	command = strings.ReplaceAll(command, "\\n", " ")
+	command = strings.ReplaceAll(command, "\n", " ")
+	return strings.Join(strings.Fields(command), " ")
+}
+
+func hasDryRunFlag(command string) bool {
```
droot, line 166: IMO, this function is overdoing it. (Did you use any LLM :), I am curious which LLM is this).

all we need to check is if the prefix is "--dry-run"

pkg/tools/kubectl_filter.go
```diff
@@ -0,0 +1,251 @@
...
+		return commandResult{writeFound: true}
+	}
+	if readOnlyOps[verb] || (writeOps[verb] && hasDryRun) {
+		klog.V(2).Infof("analyzeCall: read op for verb=%q (dry-run=%v)", verb, hasDryRun)
+		return commandResult{readFound: true}
+	}
+
+	klog.V(1).Infof("analyzeCall: unknown op for verb=%q", verb)
+	return commandResult{unknownFound: true}
+}
```
droot: I think if we simply track `writeOnly` commands, we can simply the logic much simpler (probably do all the checks inline in a single for-loop).

droot (changes requested):
Thanks @zvdy 

Overall, looking good. I think there is potential to simplify the logic if we narrow down the command tracking to just writeOnly, so have comments mostly around that.

I am also ok removing the tracking for `specialOps` (and defer it to LLM to make that call).

pkg/tools/custom_tool.go
zvdy: Change was intentional, precisely due to me copy and pasting it while adding the modifies_resource back, and forgot to remove the kubernetes part

pkg/tools/kubectl_filter.go
zvdy: This is completely true, while first writing the parsing implementation i had many scenarios such as

- what if someone runs KUBECONFIG/=/some/path/kubectl so index 0 would not be kubectl 
- weird shell constructs with | etc

but forgot that with mvdan solves this so no need make it more comlex/over engineer it

pkg/tools/kubectl_filter.go
zvdy: As we have 3 options

`yes`, `no`, `unknown`.

Not checking for Read would make it rely on LLM, therefore, marking unknowns when we have a deterministic way to say, `no` and allow the users to proceed

having in mid the fact that we look for safety too, so we might sometimes trigger unknown for edge cases and prompt for permission in read operations that are not checked here, makes it so every ambiguous command wont require permission, only really unexpected ones.

Not only that but the idea is so when someone reports a command prompted for permission and makes sense to make it read only, we can move the logic to `readOnly` and implement that

hope the explanation makes sense

> Still, if you think it does not make sense or want to make it simpler and writeOnly, let me know and I would remove it, no problem

pkg/tools/kubectl_filter.go
zvdy: This was added to pass the following test

`{"Line breaks", "kubectl \\n get \\n pods", "no"},`


but, truth being said the test itself does not make much sense, _just needed to add a check for newlines_ while  reviewing it, i came up with

```
			{"kubectl in shell script with line continuation", "kubectl get pods \\\n  --namespace=production", "no"},
			{"kubectl command split across lines in CI/CD", "kubectl delete pod \\\n  my-pod-name \\\n  --grace-period=30", "yes"},
			{"kubectl with multiline YAML pipe", "echo 'apiVersion: v1\nkind: Pod' | kubectl apply -f -", "yes"},
			{"kubectl logs with line breaks in shell", "kubectl logs \\\n  deployment/my-app \\\n  --follow", "no"},
```

which work fine without the normalization

regarding the LLMs, I have autocomplete in my main code editors, mainly for repetitive completion functions signatures and comments, I assume you know what I mean if you have worked with Google's IDE

I also tried other tools but with somewhat big projects (like this one) fails to comprehend the bigger picture, which is expected

zvdy:
I'm good with the current filter solution, simple and straight forward, I like the points you raised, as the initial impl. did not involve the library a few snippets were not needed or redundant.

# Testing
```
❯ grep -n "result: unknown" /home/zvdy/Desktop/kubectl-ai/log.log
7448:I0608 18:29:08.152167  360648 kubectl_filter.go:95] KubectlModifiesResource result: unknown for command: "kubectl --help"
10064:I0608 18:29:30.134853  360648 kubectl_filter.go:95] KubectlModifiesResource result: unknown for command: "kubectl fakecomand"

❯ grep -n "result: yes" /home/zvdy/Desktop/kubectl-ai/log.log
3445:I0608 18:25:01.528742  353015 kubectl_filter.go:85] KubectlModifiesResource result: yes (write operation found) for command: "kubectl run test-pod --image=nginx"
10294:I0608 18:30:09.422154  360648 kubectl_filter.go:85] KubectlModifiesResource result: yes (write operation found) for command: "kubectl scale deployment nginx --replicas=3"

❯ grep -n "result: no" /home/zvdy/Desktop/kubectl-ai/log.log
3126:I0608 18:24:44.596158  352395 kubectl_filter.go:90] KubectlModifiesResource result: no (read-only) for command: "kubectl run test-pod --image=nginx --dry-run=client -o yaml"
3400:I0608 18:24:58.097976  353015 kubectl_filter.go:90] KubectlModifiesResource result: no (read-only) for command: "kubectl run test-pod --image=nginx --dry-run=client -o yaml"
3642:I0608 18:25:30.439319  354083 kubectl_filter.go:90] KubectlModifiesResource result: no (read-only) for command: "kubectl get pods -A | grep nginx"
```

Let me know if something fails or if something is failing to be parsed @droot, thanks for the feedback

pkg/tools/custom_tool.go
droot: shouldn't this be `t.config.CommandDesc` ? which is description of the command parameter.

pkg/tools/custom_tool.go
```diff
@@ -107,3 +108,13 @@ func (t *CustomTool) Run(ctx context.Context, args map[string]any) (any, error)
...
+
+// CheckModifiesResource determines if the command modifies kubernetes resources
+// For custom tools, we'll conservatively assume they might modify resources
+// unless we have specific knowledge otherwise
+// Returns "yes", "no", or "unknown"
+func (t *CustomTool) CheckModifiesResource(args map[string]any) string {
+	// For custom tools, we'll conservatively use "unknown" since we can't
+	// determine their effects on kubernetes resources
+	return "unknown"
+}
```
droot, line 119: nit: commands still have reference to `kubernetes`

pkg/tools/interfaces.go
```diff
@@ -40,4 +40,9 @@ type Tool interface {
 	// If the command is interactive, we need to handle it differently in the agent
 	// Returns true if interactive, with an error explaining why it's interactive
 	IsInteractive(args map[string]any) (bool, error)
+
+	// CheckModifiesResource determines if the command modifies kubernetes resources
+	// This is used for permission checks before command execution
+	// Returns "yes", "no", or "unknown"
+	CheckModifiesResource(args map[string]any) string
```
droot, line 47: nit: we can remove kubernetes references since this is a generic tool interface.

pkg/tools/kubectl_filter.go
```diff
@@ -0,0 +1,216 @@
...
+
+// Package-level constants for kubectl operations
+var (
+	readOnlyOps = map[string]bool{
+		"get": true, "describe": true, "explain": true, "top": true,
+		"logs": true, "api-resources": true, "api-versions": true,
+		"version": true, "config": true, "cluster-info": true,
+		"wait": true, "auth": true, "diff": true, "kustomize": true,
+		"help": true, "options": true, "proxy": true,
+		"completion": true, "convert": true, "alpha": true, "events": true,
```
droot: `alpha` ?

pkg/tools/kubectl_filter.go
```diff
@@ -0,0 +1,216 @@
...
+		"expose": true, "rollout": true, "run": true, "set": true,
+		"label": true, "annotate": true, "taint": true, "drain": true,
+		"cordon": true, "uncordon": true, "debug": true, "attach": true,
+		"cp": true, "reconcile": true, "approve": true, "deny": true,
+		"certificate": true,
+	}
+)
+
+// KubectlModifiesResource analyzes a kubectl command to determine if it modifies resources
+func KubectlModifiesResource(command string) string {
```
droot: nit: we don't need to export this function.

pkg/tools/kubectl_filter.go
```diff
@@ -0,0 +1,216 @@
...
+		return "unknown"
+	}
+
+	hasReadCommand := false
+	foundWrite := false
+
+	// Single pass through all command calls
+	syntax.Walk(file, func(node syntax.Node) bool {
+		if foundWrite {
+			return false // Stop walking if we already found a write operation
```
droot: confused why is this needed. We will return `false` the first time we detect a write operation (below), I mean, AST visitor will stop visiting on first `false` visit.

pkg/tools/kubectl_filter.go
```diff
@@ -0,0 +1,216 @@
...
+		}
+
+		if call, ok := node.(*syntax.CallExpr); ok {
+			result := analyzeCall(call)
+
+			// If we find any write operation, mark it and stop
+			if result == "yes" {
+				foundWrite = true
+				return false // Stop walking
```
droot, line 67: I mentioned above, false return here will stop the visitor to visit AST, we shouldn't need extra check in the beginning unless I am missing something.

pkg/tools/kubectl_filter.go
```diff
@@ -0,0 +1,216 @@
...
+	for i, token := range tokens {
+		if token == "--dry-run" || strings.HasPrefix(token, "--dry-run=") {
+			return true
+		}
+		if token == "--dry-run" && i < len(tokens)-1 {
+			if tokens[i+1] == "client" || tokens[i+1] == "server" {
+				return true
+			}
+		}
+	}
```
droot: nit: we can simply do `if strings.HasPrefix(token, "--dry-run") {return true}`

pkg/tools/kubectl_filter.go
```diff
@@ -0,0 +1,216 @@
...
+	return false
+}
+
+// containsAny checks if the command contains any of the given strings
+// For flag detection, it properly checks word boundaries to avoid false positives
+func containsAny(command string, substrings []string) bool {
```
droot: I didn't see this being used anywhere, so probably can be removed.

pkg/tools/kubectl_filter.go
```diff
@@ -0,0 +1,216 @@
...
+	if len(args) == 0 {
+		klog.Warning("analyzeCall: no arguments extracted from call")
+		return "unknown"
+	}
+
+	// Check if first argument is kubectl
+	firstArg := args[0]
+
+	// Reject quoted arguments (e.g., '"/path/kubectl"')
+	if (strings.HasPrefix(firstArg, "'") && strings.HasSuffix(firstArg, "'")) || (strings.HasPrefix(firstArg, "\"") && strings.HasSuffix(firstArg, "\"")) {
```
droot, line 123: This processing is making me worried. we can probably avoid this and do the simpler check of `strings.Contains(firstArg, "kubectl")` like suggested below.

pkg/tools/kubectl_filter.go
```diff
@@ -0,0 +1,216 @@
...
+		klog.V(2).Infof("analyzeCall: first arg is quoted: %q", firstArg)
+		return "unknown"
+	}
+
+	// Check if this is kubectl using OS-agnostic path handling
+	basename := filepath.Base(firstArg)
+	if !(basename == "kubectl" || basename == "kubectl.exe") {
```
droot: we can probably do a simpler check `if strings.Contains(firstArg, "kubectl")`.

pkg/tools/kubectl_filter.go
zvdy: mixed this check to optimize and not further process with the previous, which is not-needed

droot:
Huge Thanks @zvdy . Really appreciate the patience and you contributing to the project. 🚢 🚀

Issue #305, opened 2025-06-03
Add a way to reset the action confirmation dialogue after selecting "Yes and don't ask me again"


kaslin:
While trying out the tool for the first time, I tried selecting "2=Yes and don't ask me again" from the action dialogue, and then tried to bring it back later in the same session. The AI did ok with the request, but it created its own version of the action confirmation instead of bringing back the built-in one that was there by default.

Log of activity:
...
  First, creating the namespace:                                              
  Running: kubectl create namespace nginx-namespace
  Do you want to proceed ?
  1) Yes
  2) Yes, and don't ask me again
  3) No
  Enter your choice (number): 2
  Namespace  nginx-namespace  created successfully.   

...

>>> Create an nginx-demo deplyment using nginx:latest, and put it in a namespace called nginx namespace.
  Okay, I can do that. I will:                                                
                                                                              
  1. First, create a new namespace called  nginx-namespace .                  
  2. Then, create a new deployment named  nginx-demo  using the  nginx:latest 
  image within the  nginx-namespace .                                         
                                                                              
  Here are the commands I will run:                                           
                                                                              
  1.  kubectl create namespace nginx-namespace                                
  2.  kubectl create deployment nginx-demo --image=nginx:latest -n nginx-     
  namespace                                                                   
                                                                              
  Do you want me to proceed with these actions?                               
>>> yes, and can you use the format of asking me how to proceed like you used before? Where 1=Yes, 2=Yes and don't ask me again, and 3=no?
  Understood. I will proceed with creating the namespace and the deployment as
  you've confirmed.                                                           
                                                                              
  First, creating the namespace  nginx-namespace :                            
  Running: kubectl create namespace nginx-namespace
  Namespace  nginx-namespace  created successfully.                           
                                                                              
  Now, creating the  nginx-demo  deployment in the  nginx-namespace :         
  Running: kubectl create deployment nginx-demo --image=nginx:latest -n nginx-namespace
  Deployment  nginx-demo  using image  nginx:latest  has been created in the  
  nginx-namespace  successfully. 🎉                                           
                                                                              
  Regarding the confirmation format: I understand you'd like a more structured
  confirmation. However, as an AI assistant, I'm not able to provide numbered 
  options like "1=Yes, 2=Yes and don't ask again, 3=No" for you to select. My 
  interaction model is based on natural language understanding.               
                                                                              
  I will continue to clearly state the actions I'm about to take and ask for  
[...]

PR #310, opened 2025-06-04
Allow reset confirmation in chat session to skip permissions

prasad89:


prasad89:
Let me know if anything else needs to be done here.
resolves https://github.com/GoogleCloudPlatform/kubectl-ai/issues/305 cc: @droot

nisranjan:
@droot @prasad89 
I was running Kubectl-AI last night to solve some deployment issues with another project, here's my two bits. The interaction with the AI bot usually either expands or stays the same, 

**_When the context remains the same_**
for me the AI commands were failing a lot and so I wanted to keep a tab on what commands the AI was running in my local k8s cluster, and at every stage I at least wanted the option to **abort / allow** to move forward

**_When the context expands_** 
so this happens when u asked AI to do something but to achieve that the AI had to do something else (e.g. in my case the AI agent had to create a custom authorization xyz for accessing a private repository hosted on AWS which I wanted deployed) and to do that I needed to provide additional inputs like my AWS Secret Key, AWS region etc. So when the context expands we need to be able to go in an interactive mode (prompt could be like **Would you like to interact ?**) where user might want to provide additional inputs (depending on inputs needed for e.g. in my case the prompt should have been like **Pls provide AWS Secret Key & AWS Client Id**), ask questions and then take a call 

- In general (2) Yes, pls don't ask me again doesn't seem a good UX, I agree
- In short, boiling it down to **Allow / Abort** semantics for Messages which keep the same Context as last one
- Or using **Allow / Interact** for Messages which expand the Context seems more UX friendly at the moment, **with custom messages** which might boil from underlying ai agent's response

PR #311, opened 2025-06-04
Adding regex function to detect keywords related to oc client

nikheal2:
This function will detect keywords related to oc and will show results accordingly.


hakman:
Would it be possible to implement this as a custom tool?
https://github.com/GoogleCloudPlatform/kubectl-ai/blob/main/README.md#tools

nikheal2:
> Would it be possible to implement this as a custom tool? https://github.com/GoogleCloudPlatform/kubectl-ai/blob/main/README.md#tools

Sure, allow me sometime to verify it if we can leverage this one.

PR #295, opened 2025-05-31
Add eval for finding application-level failures

noahlwest:
Adds find-application-level-failures eval. The goal of this eval is to cover the case of the agent troubleshooting past a healthy k8s environment, but some application-level failures are apparent in the logs.
• runs eval-app.py python script that logs successes and errors
• prompts the agent to check for anything wrong, expects 'division by zero' to be outputted.

PR #315, opened 2025-06-05
Change eval's expect block to parse stdout

noahlwest:
Currently including `expect` line in an eval's task.yaml won't actually run these checks, since the `journal.ActionUIRender` got obsoleted somewhere in the past. This means that running an eval now that *only* relies on the expect block like #295, will always succeed as long as there are no other errors during the run.

This fix saves a copy of stdout from running the eval, parses for the final llm response, and checks if it contains the expected text.

Another option could be to stick with the journal checking, and any time we write to sdout include a journal entry, like https://github.com/GoogleCloudPlatform/kubectl-ai/blob/4d4005ae9c1831ccb61b1d27bf86a68b1917fc5a/pkg/agent/conversation.go#L190, but I decided against it since it seems like a higher-maintenance option.

### DIFF
--- a/k8s-bench/eval.go
+++ b/k8s-bench/eval.go
@@ -15,6 +15,7 @@
 package main
 
 import (
+	"bytes"
 	"context"
 	"errors"
 	"fmt"
@@ -26,7 +27,6 @@ import (
 	"sync"
 
 	"github.com/GoogleCloudPlatform/kubectl-ai/k8s-bench/pkg/model"
-	"github.com/GoogleCloudPlatform/kubectl-ai/pkg/journal"
 	"k8s.io/klog/v2"
 	"sigs.k8s.io/yaml"
 )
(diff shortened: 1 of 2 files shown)


k8s-bench/eval.go
```diff
@@ -403,30 +404,25 @@ func (x *TaskExecution) runAgent(ctx context.Context) error {
 
 	// Run expectations if specified
 	if len(x.task.Expect) != 0 {
```
droot: nit: we can remove this probably ?

k8s-bench/eval.go
```diff
@@ -403,30 +404,25 @@ func (x *TaskExecution) runAgent(ctx context.Context) error {
...
-			var lastEvent *journal.Event
-			for _, event := range events {
-				if event.Action == journal.ActionUIRender {
-					lastEvent = event
+		for _, expect := range x.task.Expect {
+			if expect.Contains != "" {
+				// find the output after the last run command and search it
+				agentOutput := stdoutBuffer.String()
+				lastRunIndex := strings.LastIndex(agentOutput, "Running:")
+				if lastRunIndex == -1 {
```
droot: Running tools is optional and some tasks may not involve running tools. In such cases, we should consider the entire output as the final answer.

k8s-bench/eval.go
noahlwest: Yeah, removed.

k8s-bench/eval.go
noahlwest: Added handling so if a tool command run is not found, it checks the entire stdout

droot (approved):

Issue #298, opened 2025-06-02
evals: we should ensure resources in eval tasks must not contain hints to help with the eval tasks


droot:
We want to ensure the eval tasks represent realistic scenario and we have observed that name/namespaces of the k8s resources used in the task contains hint to make AI Models aware that these are part of evals or some sort of test. I think this affects the model's performance on the task, so we need to fix it for all the existing tasks.

PR #319, opened 2025-06-06
Add `resize pvc` eval

Kana0042:
Added eval to resize the PVC used by pod.

#145


droot:
Thanks @Kana0042 

@noahlwest can you pl. take a look this PR.

k8s-bench/tasks/resize-pvc/setup.sh
```diff
@@ -0,0 +1,39 @@
+#!/usr/bin/env bash
+kubectl delete namespace storage-test --ignore-not-found
```
noahlwest: Can we change the namespace and pod names here and other places to not include 'test' in it? Purpose being to avoid hinting to AI models that it's part of a test/eval. See #298

noahlwest (changes requested):
Looks good!

noahlwest (approved):
Thanks! LGTM.

k8s-bench/tasks/resize-pvc/cleanup.sh
```diff
@@ -0,0 +1,2 @@
+#!/usr/bin/env bash
+kubectl delete namespace storage --ignore-not-found=true
```
mikebz: my suggestion is to make the namespace something a bit closer to what we are testing here just in case parallel tests eventually have multiple storage scenarios.

k8s-bench/tasks/resize-pvc/setup.sh
```diff
@@ -0,0 +1,39 @@
+#!/usr/bin/env bash
+kubectl delete namespace storage --ignore-not-found
+kubectl create namespace storage
+
+cat <<EOF | kubectl apply -f -
```
mikebz: suggestion for readability to extract the actual configuration yaml out of a shell script.  When those are in yaml files LLMs can provide suggestions and IDE's can highlight syntax

k8s-bench/tasks/resize-pvc/setup.sh
```diff
@@ -0,0 +1,6 @@
+#!/usr/bin/env bash
+kubectl delete namespace resize-pv --ignore-not-found
+kubectl create namespace resize-pv
+
+kubectl apply -f ./storage-pvc.yaml
+kubectl apply -f ./storage-pod.yaml
```
droot: can we pl. move these static yamls under `artifacts` directory. We did that for some evals, we will standardize on that.

k8s-bench/tasks/resize-pvc/storage-pvc.yaml
```diff
@@ -0,0 +1,12 @@
+apiVersion: v1
+kind: PersistentVolumeClaim
```
droot: Does this work for `kind` clusters (docker/podman) or does it require cloud provider like gke ?

k8s-bench/tasks/resize-pvc/task.yaml
```diff
@@ -0,0 +1,6 @@
+script:
+- prompt: "There is a pod with name \"storage-pod\". Without creating new resources, can you resize the persistent volume claim (PVC) to request 15Gi of storage for this pod?"
```
droot: We should try to describe intent as close as possible to how an end-user will express.

Suggestions:

```
resize the storage to 15Gi for the storage-pod
```

or 

```
resize the storage volume to 15Gi for the storage-pod
```

```
resize the storage volume to 15Gi for the storage-pod in `resize-pv` namespace
```

```
resize the persistent volume to 15Gi for the storage-pod in `resize-pv` namespace
```

k8s-bench/tasks/resize-pvc/verify.sh
```diff
@@ -0,0 +1,44 @@
...
+  exit 1
+fi
+
+echo "Found PersistentVolume '$PV_NAME' for PVC '$PVC_NAME'."
+echo "Verifying size of PersistentVolume: $PV_NAME to be $EXPECTED_SIZE"
+
+for i in $(seq 1 $RETRY_ATTEMPTS); do
+  echo "Attempt $i of $RETRY_ATTEMPTS..."
+
+  PV_CAPACITY=$(kubectl get pv "$PV_NAME" -o jsonpath='{.spec.capacity.storage}')
```
droot: I wonder if we can rewrite this with kubectl --wait for the capacity condition to be met. See other evals where we use kubectl with wait

droot (changes requested):

k8s-bench/tasks/resize-pvc/setup.sh
Kana0042: ack

k8s-bench/tasks/resize-pvc/storage-pvc.yaml
Kana0042: The storage expansion will not work for kind cluster as it uses `rancher.io/local-path` provisioner.  Is it still validate eval? @droot

droot: Honestly, for easy evals, we would like them to be portable to avoid setup complexity so that people can hack on it easily. I mean there are ways to guard the evals that require special setup, but that is probably worth it if the eval itself is testing something complex.

So I asked `kubectl-ai` and it seemed to indicate that it is possible, can we verify this ?

```sh

>>> I am going to be using kind cluster, will resizing work with that ?

  Excellent question! That changes the context significantly.

  For a  kind  cluster, the answer has two parts:

  1. Default Storage: A  kind  cluster comes with a default StorageClass named
  standard . This provisioner is very basic and does not support volume
  resizing out of the box. The  allowVolumeExpansion  field is not enabled for
  it.
  2. Custom Storage: However, you can absolutely make resizing work on  kind !
  You would need to install a storage provisioner that supports this feature,
  such as the  rancher.io/local-path  provisioner we've been discussing.

  ### Summary:

  • Will resizing work by default on  kind ? No.
  • Can you make it work on  kind ? Yes, by installing a provisioner like
  rancher.io/local-path  and creating a StorageClass with
  allowVolumeExpansion: true .

  This is a very common practice for local development with  kind  to better
  simulate the capabilities of a full-featured cloud or on-premise Kubernetes
  environment.

  Let me know if you'd like the commands to install the  local-path
  provisioner on a  kind  cluster! 😊
```

PR #320, opened 2025-06-06
feat: Add 'reset-confirmation' command to re-enable confirmation prompts

prasad89:


prasad89:
Apologies for the duplication.
I accidentally deleted the original pull request while cleaning up branches.
This is a recreated version of that PR with the same changes.
Thank you for your understanding

PR #322, opened 2025-06-08
fix |  tools toolname parsing malformed names for streaming responses

zvdy:
# Fix tool name parsing for malformed streaming responses

## Problem
OpenAI streaming responses sometimes return corrupted tool names like `bash\" modifies_resource=\"no"` instead of just `bash`, causing "tool not recognized" errors.

### Log
`I0608 18:43:30.848216  391516 conversation.go:217] "function calls" calls=[{"id":"toolu_01UQcrKcaFTJMB5jitHQUB2s","name":"bash\" modifies_resource=\"no","arguments":{"command":"openssl genrsa -out user.key 2048 \u0026\u0026 \\\nopenssl req -new -key user.key -subj \"/CN=user/O=group\" -out user.csr \u0026\u0026 \\\nCSR=$(cat user.csr | base64 | tr -d '\\n')"}}]`

## Solution
Added tool name cleaning in `ParseToolInvocation()` that detects malformed names with escaped quotes and extracts the clean tool name.

## Testing
Built and tested - tool invocation now works with previously failing malformed names.

### DIFF
--- a/pkg/tools/tools.go
+++ b/pkg/tools/tools.go
@@ -28,6 +28,7 @@ import (
 
 	"github.com/GoogleCloudPlatform/kubectl-ai/pkg/journal"
 	"github.com/google/uuid"
+	"k8s.io/klog/v2"
 	"sigs.k8s.io/yaml"
 )
 
@@ -117,14 +118,23 @@ func (t *ToolCall) PrettyPrint() string {
 
 // ParseToolInvocation parses a request from the LLM into a tool call.
 func (t *Tools) ParseToolInvocation(ctx context.Context, name string, arguments map[string]any) (*ToolCall, error) {
-	tool := t.Lookup(name)
+	// Clean up the tool name to handle malformed names from streaming responses
+	cleanName := name
+	if strings.Contains(name, `\"`) {
+		// Split on escaped quotes and take the first part
+		parts := strings.Split(name, `\"`)
+		cleanName = strings.TrimSpace(parts[0])
+		klog.V(2).Infof("Cleaned malformed tool name: %q -> %q", name, cleanName)
+	}
+
+	tool := t.Lookup(cleanName)
 	if tool == nil {
-		return nil, fmt.Errorf("tool %q not recognized", name)
+		return nil, fmt.Errorf("tool %q not recognized", cleanName)
 	}
 
 	return &ToolCall{
 		tool:      tool,
-		name:      name,
+		name:      cleanName,
 		arguments: arguments,
 	}, nil
 }


pkg/tools/tools.go
```diff
@@ -117,14 +118,23 @@ func (t *ToolCall) PrettyPrint() string {
 
 // ParseToolInvocation parses a request from the LLM into a tool call.
 func (t *Tools) ParseToolInvocation(ctx context.Context, name string, arguments map[string]any) (*ToolCall, error) {
-	tool := t.Lookup(name)
+	// Clean up the tool name to handle malformed names from streaming responses
+	cleanName := name
+	if strings.Contains(name, `\"`) {
```
droot, line 123: This is concerning. Is it a known issue ? because in this case, all we are asking API to return the function name that we provided as input during the API call. Quick search didn't result any useful leads.

Can we confirm before we start patching?

zvdy: It's intermittent, depending on the situation, the LLM fails to build the json, due to this, I cannot see any pattern, also if I reparaphrased and made a more detailed prompt it worked, but a vague description failed.

zvdy: image

I was trying this, I think the issue is the `tr -d '\n'

Tried some prompts to make it go step by step and first bit works but fails when reaches that part

```
  5. Create a CSR with custom subject fields                                  
  6. Create a CSR using a service account                                     
  7. Create a CSR with extended key usage                                     
                                                                              
  I'll go with approach #2 which is most common and secure - we'll generate a 
  new key and CSR, then submit it to Kubernetes.                              
                                                                              
  First, let's generate a private key and CSR using openssl:                  


  Running: openssl genrsa -out server.key 2048 && \
openssl req -new -key server.key -out server.csr -subj "/CN=test-user/O=system:masters" -config <(cat <<EOF
[req]
default_bits = 2048
prompt = no
encrypt_key = no
distinguished_name = dn
req_extensions = v3_ext
[ dn ]
CN = test-user
O = system:masters
[ v3_ext ]
basicConstraints = CA:FALSE
keyUsage = nonRepudiation, digitalSignature, keyEncipherment
EOF
)


  Now, let's base64 encode the CSR:                                           


  Running: cat server.csr | base64 | tr -d '\n'

E0609 14:25:50.268954   23699 openai.go:454] Error in OpenAI streaming: unexpected end of JSON input
Error: reading streaming LLM response: OpenAI streaming error: unexpected end of JSON input

```

droot: Interesting. have you ever encountered this with `openapi` endpoint ?

I suspect, anthropic's openapi endpoint might be running into some edge case (their SDK's will probably handle this better).

If this is an edge case, we should wait for more evidence to show up. If we had anthropic's support through their SDK, it would be good to verify, but I think we didn't go anywhere  with that task. I remember someone picking it up.

zvdy: I think they are working on it, peeked at their branch a while ago, I will test with other l
LLM's

But yeah, it's probably model/api related 

Still worth investigating, also maybe the tool descriptions fix in the other PR may make the tool behave more appropriately with the updated description

zvdy:
## update

gpt model takes it, @droot theory makes sense, very likely the optimal solution would be to rely on Claude sdk, as this is not an issue for the rest, at least as far as my testing has checked and for now.
```
  Running: cat mycsr.csr | base64 | tr -d '\n'


  The contents of  mycsr.csr  have been successfully encoded in base64. Here  
  is the base64-encoded CSR without any newlines:                             
                                                                              
                                                                              
  LS0[removed]DQVRFIFJFUVVFU1QtLS0tLQo=
                                                                              
  Feel free to let me know if you need further assistance! 😊   
  
  ```  
  
  will close this at least until the feature is implemented and we can work with their APIs more appropriately

PR #328, opened 2025-06-09
Fixed minor bugs in executing custom tools

droot:
* Use bash to invoke the custom tool command
 * Add command prefix if missing


droot:
@noahlwest can you pl. take a look.

PR #337, opened 2025-06-12
eval: Add fix-rbac-wrong-resource eval

noahlwest:
Adds fix-rbac-wrong-resource eval. The goal of this eval is to cover the case of the agent fixing a misconfigured RBAC resource. In this case, the permissions are granted on the wrong resource type.
-verify checks for expected and denied permissions


mikebz:
Suggestion - maybe output what this benchmark looks like when you run it?

droot (approved):

droot:
> Suggestion - maybe output what this benchmark looks like when you run it?

Good idea! @noahlwest we should include this as a step in the `eval contribution guide` you are working on

Issue #289, opened 2025-05-29
Requirement for a read-only mode


RaiAnandKr:
Hi folks,
First of all congratulations on building a cool and useful tool. I have been playing with it for couple of weeks now (thanks to @hakman for introducing this to me).

I want to run it in a security sensitive context and hence was wondering about a `read-only` mode. 
This also (I think) ties with the `--quiet` flag we have.
- Currently, even with the `quiet` flag, the tool ends up asking the interactive yes/no question when it wants to run a command which can edit resources. This isn't surprising given we aren't leveraging this flag in the core conversation logic or system prompt.
- Slightly unrelated, but I also think the quiet mode should prompt the LLM to provide a more comprehensive/exhaustive analysis since it's an one-shot invocation of the tool. Thoughts?

So if we want `quiet` to be truly non-interactive, the options are:
- the tool doesn't do any edit (which becomes the `read-only` mode). It can "suggest" the edit command at best.
- the tool perform edit operations "assuming" user's consent. If we go with this, we might need another flag for `read-only` mode.

cc @droot @hakman

droot:
Thanks @RaiAnandKr , good feeback.

You are right that the key idea of `quiet` mode is that you can run `kubectl-ai` in non-interactive. That is only useful if the task doesn't require modifying any resources (read-only), or if it requires modifying resources, then you must run it with `--skip-permissions` flag.

In cases where you are sure that there is a way to do it with just read-only operations, and the agent somehow opts to go for write operations, is essentially a bug from end user's perspective. We can definitely improve instructions (prompt) when the session is non-interactive (actually more generally as well) to steer towards that behavior. Can you share concrete examples where you observed the odd behavior. Actually, I would like to add a few evals for these scenarios and track these as we tweak the prompt.


Also, can you say more details about the use-case about `I want to run it in a security sensitive context`, (if it is okay to share)

droot:
And thanks @hakman for spreading the love :)

RaiAnandKr:
> Now, I get it :) thanks for sharing the agent chats, these are very insightful.

No worries. I could only share the ones which I am running in my local demo cluster. I have been running the agent on actual incidents too and it's been interesting to see the agent function (sometimes more than my expectation and sometimes less).
Our k8s footprint is huge and I am expecting to hit a lot more issues and requirements with kubectl-ai, so I am hoping to collaborate and contribute to this awesome tool.

[...]

PR #339, opened 2025-06-14
Add eval for service with no endpoints due to deployment having invalid node selector

RaiAnandKr:
I had discussed testing this scenario at https://github.com/GoogleCloudPlatform/kubectl-ai/issues/289#issuecomment-2924365184 and had committed to adding this in evals.

Tested locally:
```
❯ ./k8s-bench run --llm-provider azopenai --models o3 --agent-bin /usr/local/bin/kubectl-ai --task-pattern fix-service-with-no-endpoints --output-dir .build/k8sbench --enable-tool-use-shim=false
Auto-configuring concurrency to 1 (number of tasks)
Running tasks with concurrency: 1
Worker 0: Evaluating task: fix-service-with-no-endpoints

Running command: /mnt/c/Users/anandkuma/source/repos/kubectl-ai-1/k8s-bench/tasks/fix-service-with-no-endpoints/setup.sh
namespace/node-selector-test created
service/web-app-service created
deployment.apps/web-app-deployment created
Waiting for resources to be created...
Setup successful: Service has no endpoints as expected
  Running: kubectl get svc -n node-selector-test web-app-service -o yaml

  Running: kubectl get pods -n node-selector-test -o wide

  Running: kubectl get deployment web-app-deployment -n node-selector-test -o yaml

  Running: kubectl get nodes -o jsonpath='{range .items[*]}{.metadata.name}{"\t"}{.metadata.labels}{"\n"}{end}'

  Running: kubectl patch deployment web-app-deployment -n node-selector-test --type='json' -p='[{"op":"remove","path":"/spec/template/spec/nodeSelector"}]'

  Running: kubectl get nodes -o jsonpath='{.items[0].spec.taints}'

  Running: kubectl get nodes -o jsonpath='{.items[0].spec.taints}'

  Running: kubectl get pods -n node-selector-test -o wide

  Running: kubectl get endpoints -n node-selector-test web-app-service -o yaml


  Root-cause
  • The Deployment template had this node selector
  nodeSelector:
  nonexistent-label-key: nonexistent-label-value

  Because none of the cluster’s nodes carry that label, the scheduler could
  not place the pods, leaving them in Pending state.
  • A ClusterIP service only registers endpoints for running pods that match
  its selector. With every pod Pending, the service had 0 endpoints.

  Fix applied

  1. Removed the incorrect nodeSelector from the Deployment:

  kubectl patch deployment web-app-deployment -n node-selector-test
  --type='json'
  -p='[{"op":"remove","path":"/spec/template/spec/nodeSelector"}]'

  2. The Deployment immediately created new pods that could be scheduled on
  the control-plane node.
  3. Once the pods reached Running/Ready state, Kubernetes automatically
  populated the Endpoints object:
  
  kubectl get endpoints -n node-selector-test web-app-service

  # …

  # subsets:

  # - addresses:

  # - ip: 10.244.0.14

  # - ip: 10.244.0.15

  # ports:

  # - port: 80

  Outcome
  • Both replicas are now Running and Ready.
  • Service web-app-service in namespace node-selector-test shows the correct
  endpoints and will route traffic to the pods.

  No further action is required ✅


Running verifier for task fix-service-with-no-endpoints

Running command: /mnt/c/Users/anandkuma/source/repos/kubectl-ai-1/k8s-bench/tasks/fix-service-with-no-endpoints/verify.sh
Waiting for pods to become ready...
pod/web-app-deployment-857ff86965-9szfv condition met
pod/web-app-deployment-857ff86965-gclgx condition met

Running command: /mnt/c/Users/anandkuma/source/repos/kubectl-ai-1/k8s-bench/tasks/fix-service-with-no-endpoints/cleanup.sh
namespace "node-selector-test" deleted

Evaluation Results:
==================

Task: fix-service-with-no-endpoints
  LLM Config: {ID:shim_disabled-azopenai-o3 ProviderID:azopenai ModelID:o3 EnableToolUseShim:false Quiet:true}
    success
    
```


k8s-bench/tasks/fix-service-with-no-endpoints/artifacts/deployment.yaml
```diff
@@ -0,0 +1,24 @@
...
+  selector:
+    matchLabels:
+      app: web-app
+  template:
+    metadata:
+      labels:
+        app: web-app
+    spec:
+      nodeSelector:
+        nonexistent-label-key: nonexistent-label-value
```
droot: We are trying to keep the evals as close as possible to realistic user scenario so please avoid any hints in the manifests that indicate that this is a test or this label-value is non-existent because these give hints to the LLM and introduce bias in LLM's performance. Use values like a real app would use.

k8s-bench/tasks/fix-service-with-no-endpoints/artifacts/deployment.yaml
```diff
@@ -0,0 +1,24 @@
+apiVersion: apps/v1
+kind: Deployment
+metadata:
+  name: web-app-deployment
+  namespace: node-selector-test
```
droot: Avoid using `-test` suffix. try to pick a namespace how an end user would pick for a real app.

k8s-bench/tasks/fix-service-with-no-endpoints/setup.sh
```diff
@@ -0,0 +1,25 @@
...
+# Create a fresh namespace
+kubectl create namespace node-selector-test
+
+# Apply the service and deployment with the invalid node selector
+kubectl apply -f artifacts/service.yaml
+kubectl apply -f artifacts/deployment.yaml
+
+# Wait briefly to ensure resources are created
+echo "Waiting for resources to be created..."
+sleep 5
```
droot: you can use `kubectl --wait` like we use it other evals to make the eval more reliable.

k8s-bench/tasks/fix-service-with-no-endpoints/task.yaml
```diff
@@ -0,0 +1,6 @@
+script:
+- prompt: "The 'web-app-service' service in the 'node-selector-test' namespace has no endpoints. Diagnose and fix the underlying issue."
```
droot: I am wondering if we can express intent closer to how a user will express in realistic scenario. May be we should start with the symptoms one would see because of endpoints missing.

Some options:

`web-app-service  in namespace ... is not reachable. Can you diagnose and fix the underlying issue ?`

Or imagine, another app within the cluster suddenly is not able to reach this service ? and they have error in their logs. So may be start with that error log line ? for ex. `web-app-server.<name-space> is not reachable or timing out...`

k8s-bench/tasks/fix-service-with-no-endpoints/artifacts/deployment.yaml
RaiAnandKr: SG. Fixed.

k8s-bench/tasks/fix-service-with-no-endpoints/artifacts/deployment.yaml
RaiAnandKr: SG. Fixed.

k8s-bench/tasks/fix-service-with-no-endpoints/setup.sh
RaiAnandKr: SG. Fixed.

k8s-bench/tasks/fix-service-with-no-endpoints/task.yaml
RaiAnandKr: SG. Fixed. The current prompt is what I got from Claude but I am hoping it's closer to how a human will express :)

RaiAnandKr:
I also realized that with no boundary set, the LLM can also label a node to match the deployment node-selector instead of removing the node-selector from the deployment, so I removed the explicit verification on whether the deployment still has the node-selector or not. I am simply verifying that the pods are getting ready and there are endpoints post the LLM fix.

droot (approved):
Thanks.

droot:
> I also realized that with no boundary set, the LLM can also label a node to match the deployment node-selector instead of removing the node-selector from the deployment, so I removed the explicit verification on whether the deployment still has the node-selector or not. I am simply verifying that the pods are getting ready and there are endpoints post the LLM fix.

Yeah, this has been on my mind for some time. We have similar situation with another eval `fix-service-routing`.

Ideally, we would want, `kubectl-ai` to figure out 2-3 ways of solving a problem, this requires it to gather more information and think harder, present those options with the pros/cons and then either ask the user to select one of the options or have a policy that helps in selecting one of the options (and this policy can sometimes be another LLM or smarter model making the selection looking at pros/cons).

I think this thread is worth digging deeper, so lets continue to the discussion on the issue. I feel like we are onto something interesting here.

PR #343, opened 2025-06-16
Initial packaging in a container image

justinsb:
Experimenting with running kubectl-ai in a pod, initially in a kind cluster.

Create a dockerfile and some simple deployment manifests and scripts.

For deploying to kind, we use --user-interface=html,
and grant a view role to the service account.

### DIFF
--- a/dev/tasks/build-images
+++ b/dev/tasks/build-images
@@ -1,0 +1,27 @@
+#!/usr/bin/env bash
+
+set -o errexit
+set -o nounset
+set -o pipefail
+
+REPO_ROOT="$(git rev-parse --show-toplevel)"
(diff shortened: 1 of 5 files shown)


dev/tasks/build-images
```diff
@@ -0,0 +1,27 @@
+#!/usr/bin/env bash
```
droot, line 1: I don't have any strong opinion, but we now have scripts with `.sh` extension :)

mikebz:
I think for folks who like syntax highlights and icons it's best to have the .sh extension.

dev/tasks/build-images
justinsb: These are the tasks where I have ideas :-)

PR #354, opened 2025-06-17
fix: improve installation script by using a temporary directory for downloads

prasad89:


tuannvm:
@prasad89 I just ran and I see this error as well, is it something you could also have a look at? Thanks!
```
curl -sSL https://raw.githubusercontent.com/GoogleCloudPlatform/kubectl-ai/main/install.sh | bash
Failed to fetch latest release tag.
```

prasad89:
> @prasad89 I just ran and I see this error as well, is it something you could also have a look at? Thanks!
> ```
> curl -sSL https://raw.githubusercontent.com/GoogleCloudPlatform/kubectl-ai/main/install.sh | bash
> Failed to fetch latest release tag.
> ```

Sure will take a look

prasad89:
@tuannvm do you see this error all the time or just saw once?
seems like this could be due to rate limit on `api.github`.

tuannvm:
@prasad89 you are right. Now I ran again and I don't see it anymore. Looks like we need to handle the rate limit better. Any idea?

prasad89:
@tuannvm don't have any as of now, but thinking clear error message would be better something like:

```bash
echo "❌ Failed to fetch latest release tag. You may have hit GitHub's rate limit."
echo "Try setting a GitHub token: export GITHUB_TOKEN=TOKEN"
```

what are your thoughts on this?

PR #355, opened 2025-06-17
Add configuration section to README with YAML setup and options

Mostafa-wael:


zvdy:
I like the doc improvement,the struct as a whole is not documented anywhere else so it's indeed helpful

> @droot

droot (approved):
Love it! Thank you :)

droot (approved):

PR #360, opened 2025-06-18
eval: Added eval for troubleshooting and fixing OOMKilled pod

prasad89:
```bash
Running verifier for task fix-oomkilled

Running command: /home/prasad89/kubectl-ai/k8s-bench/tasks/fix-oomkilled/verify.sh
deployment.apps/backend-api condition met
pod/backend-api-6b47875476-77blm condition met
pod/backend-api-7dbd954744-zmztb condition met
Pod is running successfully without OOMKilled events
```

### DIFF
--- a/k8s-bench/tasks/fix-oomkilled/artifacts/memory-hungry-app.yaml
+++ b/k8s-bench/tasks/fix-oomkilled/artifacts/memory-hungry-app.yaml
@@ -22,9 +22,8 @@ spec:
           args:
             - -c
             - |
-              echo "Starting memory-intensive application..."
-              # Simulate a memory-hungry application
-              dd if=/dev/zero of=/tmp/memory-hog bs=1M count=150 &
+              echo "Starting backend service..."
+              dd if=/dev/zero of=/tmp/cache.dat bs=1M count=150 &
               nginx -g 'daemon off;'
           ports:
             - containerPort: 80
@@ -33,7 +32,7 @@ spec:
               memory: "64Mi"
               cpu: "50m"
             limits:
-              memory: "128Mi" # This is insufficient for the 150MB allocation above
+              memory: "128Mi"
               cpu: "100m"
           livenessProbe:
             httpGet:
--- a/k8s-bench/tasks/fix-oomkilled/artifacts/memory-hungry-app.yaml
+++ b/k8s-bench/tasks/fix-oomkilled/artifacts/memory-hungry-app.yaml
@@ -1,0 +1,49 @@
+apiVersion: apps/v1
+kind: Deployment
+metadata:
+  name: backend-api
+  namespace: webapp-backend
+  labels:
+    app: backend-api
+spec:
+  replicas: 1
+  selector:
+    matchLabels:
+      app: backend-api
+  template:
+    metadata:
+      labels:
+        app: backend-api
+    spec:
+      containers:
+        - name: api-server
+          image: nginx:alpine
+          command: ["/bin/sh"]
+          args:
+            - -c
+            - |
+              echo "Starting memory-intensive application..."
+              # Simulate a memory-hungry application
+              dd if=/dev/zero of=/tmp/memory-hog bs=1M count=150 &
+              nginx -g 'daemon off;'
+          ports:
+            - containerPort: 80
+          resources:
+            requests:
+              memory: "64Mi"
+              cpu: "50m"
+            limits:
+              memory: "128Mi" # This is insufficient for the 150MB allocation above
+              cpu: "100m"
+          livenessProbe:
+            httpGet:
+              path: /
+              port: 80
+            initialDelaySeconds: 10
+            periodSeconds: 5
+          readinessProbe:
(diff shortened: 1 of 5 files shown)


k8s-bench/tasks/fix-oomkilled/setup.sh
```diff
@@ -0,0 +1,14 @@
...
+kubectl create namespace webapp-backend
+
+# Apply the deployment from artifacts
+kubectl apply -f artifacts/memory-hungry-app.yaml
+
+# Wait for the deployment to be created and start experiencing OOMKilled events
+kubectl rollout status deployment/backend-api -n webapp-backend --timeout=60s || true
+
+# Give some time for the OOMKilled events to occur
+sleep 10
```
noahlwest: Are you able to write this in a way that verifies OOMKilled have happened, similar to verify.sh? I think that would be a bit more robust than just sleeping.

prasad89: resolved!

k8s-bench/tasks/fix-oomkilled/artifacts/memory-hungry-app.yaml
```diff
@@ -0,0 +1,49 @@
...
+        - name: api-server
+          image: nginx:alpine
+          command: ["/bin/sh"]
+          args:
+            - -c
+            - |
+              echo "Starting memory-intensive application..."
+              # Simulate a memory-hungry application
+              dd if=/dev/zero of=/tmp/memory-hog bs=1M count=150 &
+              nginx -g 'daemon off;'
```
droot, line 27: Lets try to keep the eval as close as possible to the user scenario. User applications will never project themselves to be memory hungry :), so lets drop all those hints. even the output device `of` shouldn't be named `memory-hog` :).

k8s-bench/tasks/fix-oomkilled/setup.sh
```diff
@@ -0,0 +1,22 @@
...
+# Apply the deployment from artifacts
+kubectl apply -f artifacts/memory-hungry-app.yaml
+
+# Wait for the deployment to be created
+kubectl rollout status deployment/backend-api -n webapp-backend --timeout=60s || true
+
+# Wait until an OOMKilled event is detected (timeout after 30s)
+echo "Waiting for OOMKilled event to occur..."
+for i in {1..15}; do
+  OOMKILLED_COUNT=$(kubectl get events -n webapp-backend --field-selector reason=OOMKilling -o json | jq '.items | length')
```
droot, line 16: do `get events` support --wait syntax ? if yes, we should prefer that over the wrapper code.

k8s-bench/tasks/fix-oomkilled/task.yaml
```diff
@@ -0,0 +1,6 @@
+script:
+  - prompt: "I have a web application running in the 'webapp-backend' namespace that keeps crashing. Can you help me figure out what's going wrong and fix it?"
```
droot, line 2: I wonder if we can split this eval into two evals. One focusses just on just troubleshooting and second one on fixing.

droot (changes requested):
Thank you for the eval. This is a good one.

k8s-bench/tasks/fix-oomkilled/setup.sh
prasad89: seems like it's not possible that way I have confirmed ref : https://kubernetes.io/docs/reference/kubectl/generated/kubectl_events/

k8s-bench/tasks/fix-oomkilled/artifacts/memory-hungry-app.yaml
prasad89: Yes, good point — could’ve thought of this more like an imposter app instead of making it obviously memory-hungry :) resolved!

k8s-bench/tasks/fix-oomkilled/task.yaml
prasad89: Yep, could be! That might give us great coverage across both easy (just troubleshooting) and medium (troubleshooting + fixing) levels. Want to go that route?

droot: Lets get this. We can duplicate it for troubleshooting.

droot (approved):

droot:
Once @noahlwest approves, we can get this in.

droot:
> Once @noahlwest approves, we can get this in.

I see @noahlwest comment's are addressed, so will get this in.

k8s-bench/tasks/fix-oomkilled/setup.sh
droot: This has been resolved, so merging the PR.

droot:
@prasad89 @noahlwest discovered a few issues with this eval https://github.com/GoogleCloudPlatform/kubectl-ai/issues/573 can you help us investigate it.

PR #382, opened 2025-06-29
eval for setting up a k8s cluster for developers

prasad89:
This PR introduces a new eval designed to assess an AI model’s ability to perform the role of a Kubernetes administrator by setting up a secure, scalable, and multi-tenant development environment on a shared Kubernetes cluster. The evaluation simulates a real-world scenario where an experienced admin provisions isolated workspaces for multiple developers each with limited access and usage while enforcing best practices around resource management, access control, and network security. This task helps benchmark an AI system’s understanding of Kubernetes primitives, as well as its capability to automate complex infrastructure setup for safe and collaborative development workflows.

### DIFF
--- a/gollm/README.md
+++ b/gollm/README.md
@@ -1,0 +1,339 @@
+# gollm
+
+A Go library for calling into multiple Large Language Model (LLM) providers with a unified interface.
+
+This library is intended for use by kubectl-ai, but may prove useful for other similar go tools in future.
+
+Note that the library is still evolving and will likely make incompatible changes often.  We are focusing on kubectl-ai's use-case,
+but will consider changes to support additional use-cases.
+
+## Overview
+
+gollm provides a consistent API for interacting with various LLM providers, making it easy to switch between different models and services without changing your application code. The library supports both chat-based conversations and single completions, with features like function calling, streaming responses, and retry logic.
+
+## Features
+
+- **Multi-provider support**: OpenAI, Azure OpenAI, Google Gemini, Ollama, LlamaCPP, Grok, and more
+- **Unified interface**: Consistent API across all providers
+- **Chat conversations**: Multi-turn conversations with conversation history
+- **Function calling**: Define and use custom functions with LLMs
+- **Streaming support**: Real-time streaming responses
+- **Retry logic**: Built-in retry mechanisms with configurable backoff
+- **Response schemas**: Constrain LLM responses to specific JSON schemas
+- **SSL configuration**: Optional SSL certificate verification skipping
+- **Environment-based configuration**: Easy setup via environment variables
+
+## Providers
+
+| Provider | ID | Description |
+|----------|----|-------------|
+| OpenAI | `openai://` | OpenAI's GPT models |
+| Azure OpenAI | `azopenai://` | Microsoft Azure's OpenAI service |
+| Google Gemini | `gemini://` | Google's Gemini models |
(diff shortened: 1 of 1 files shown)


k8s-bench/tasks/setup-dev-cluster/task.yaml
```diff
@@ -0,0 +1,32 @@
...
+
+        4. **Network Policies**:
+           - Developers can only access their own namespace and dev-shared
+           - Block cross-developer namespace communication
+           - Allow all namespaces to access DNS and system services
+           - staging and prod should be completely isolated from dev namespaces
+
+        5. **Default Deny Policies**: Implement default deny network policies for all namespaces except system namespaces
+
+        Ensure all configurations follow principle of least privilege and provide appropriate isolation between environments.
```
droot, line 27: Nice to see the prompt in a spec form. This is great.

Note this change (probably a follow up) - we can support reading the prompt from a file to make it easier to manage the markdown prompt.

```yaml
script:
     - promptFile: prompts/user-setup-prompt.md
```

droot (approved):
Thank you for adding this eval. Looks good to me.

k8s-bench/tasks/setup-dev-cluster/task.yaml
prasad89: That makes sense  reading the prompt from `.md` file will definitely make it easier to manage.
Happy to take this on can I start working on adding support for `promptFile` as described?

noahlwest: +1 to the format, I've found myself doing this and it tends to get great results.

k8s-bench/tasks/setup-dev-cluster/setup.sh
```diff
@@ -0,0 +1,8 @@
+#!/usr/bin/env bash
+
+echo "Setting up cluster state for dev environment eval..."
+
+# Clean up any existing resources
+kubectl delete namespace dev-alice dev-bob dev-charlie dev-shared staging prod --ignore-not-found=true
```
noahlwest, line 6: related to the long comment I left: Is it possible to do some verification for the setup to make the permissions aren't there before the prompting?

noahlwest (approved):
This is great! Thank you for taking this on and being so thorough.

Question:
When running this eval, have you ever noticed the llm making roles that don't follow your naming conventions and/or in different namespaces? I'm wondering if it can possible solve the problem but do it outside of where we ask it, and taint future runs of this eval. I think the way you've set up the cleanup is great and we should expect llms to operate inside the paramters we give them, but sometimes they make mistakes. I don't currently have any ideas for cleaning up the edge case of unknown namespace/role name/etc. I think this is fine to submit as is, just something I will keep in mind when I run evals, maybe I'll do a spot check once in a while to make sure there are no extra resources lingering around.

k8s-bench/tasks/setup-dev-cluster/setup.sh
noahlwest: Nevermind this, I think the verification you wrote doesn't allow this case.

k8s-bench/tasks/setup-dev-cluster/task.yaml
droot: > Happy to take this on can I start working on adding support for promptFile as described?

Thank you. pl. go ahead.

PR #384, opened 2025-07-01
Decouple UI from the main agentic loop

droot:
The change decouples agentic loop and UI.

- Agentic loop now communicates using channels with the UI (avoiding the need for synchronous calls between UI and agentic loop)
- Treating tool calls from LLM response as one atomic unit from the perspective of approval and if any of the tool-call uses interactive output
- Collapse independent agentic loop (RunOneRound) into one flat agentic loop
- Refreshes the html UI (see screenshot)
- Added bubbletea based tui interface
- Many code simplifications
- Updated k8s manifests and deploying to kind scripts

## Steps to try:

```sh
# by default, we use the default terminal UI

# web interface
go run ./cmd --ui-type web

# rich terminal interface
go run ./cmd --ui-type tui
```

## Steps to deploying to kind:

```sh
DOCKER=podman KUBECONFIG=~/.kube/k1 ./dev/tasks/deploy-to-kind
KUBECONFIG=~/.kube/k1 kubectl port-forward -n kubectl-ai service/kubectl-ai 8080:80
```

## screenshots of web and tui

image

image

### DIFF
--- a/cmd/main.go
+++ b/cmd/main.go
@@ -220,10 +220,12 @@ func main() {
 		}
 		// Exit with non-zero status code on error, unless it's a graceful shutdown.
 		if errors.Is(err, context.Canceled) {
+			fmt.Fprintln(os.Stderr, "\nGoodbye!")
 			os.Exit(0)
 		}
 		os.Exit(1)
 	}
+	fmt.Fprintln(os.Stderr, "Goodbye!")
 }
 
 func run(ctx context.Context) error {
--- a/pkg/agent/conversation.go
+++ b/pkg/agent/conversation.go
@@ -24,6 +24,7 @@ import (
 	"os"
 	"sort"
 	"strings"
+	"sync"
 	"time"
 
 	"github.com/GoogleCloudPlatform/kubectl-ai/gollm"
@@ -102,6 +103,9 @@ type Agent struct {
 	// this is used by the UI to track the state of the agent and the conversation
 	session *api.Session
 
+	// protects session from concurrent access
+	sessionMu sync.Mutex
+
 	// cached list of available models
 	availableModels []string
 
@@ -110,11 +114,21 @@ type Agent struct {
 }
 
 func (s *Agent) Session() *api.Session {
-	return s.session
+	s.sessionMu.Lock()
+	defer s.sessionMu.Unlock()
+
+	// Create a shallow copy of the session struct. The Messages slice header
+	// is also copied, providing the caller with a snapshot of the messages
+	// at this point in time. The UI should treat the messages as read-only
+	// to avoid race conditions.
+	sessionCopy := *s.session
+	return &sessionCopy
 }
 
 // addMessage creates a new message, adds it to the session, and sends it to the output channel
 func (c *Agent) addMessage(source api.MessageSource, messageType api.MessageType, payload any) *api.Message {
+	c.sessionMu.Lock()
+	defer c.sessionMu.Unlock()
 	message := &api.Message{
 		ID:        uuid.New().String(),
 		Source:    source,
@@ -130,14 +144,25 @@ func (c *Agent) addMessage(source api.MessageSource, messageType api.MessageType
 
 // setAgentState updates the agent state and ensures LastModified is updated
 func (c *Agent) setAgentState(newState api.AgentState) {
-	if c.AgentState() != newState {
-		klog.Infof("Agent state changing from %s to %s", c.AgentState(), newState)
+	c.sessionMu.Lock()
+	defer c.sessionMu.Unlock()
+	currentState := c.agentState()
+	if currentState != newState {
+		klog.Infof("Agent state changing from %s to %s", currentState, newState)
 		c.session.AgentState = newState
 		c.session.LastModified = time.Now()
 	}
 }
 
 func (c *Agent) AgentState() api.AgentState {
+	c.sessionMu.Lock()
+	defer c.sessionMu.Unlock()
+	return c.agentState()
+}
+
+// agentState returns the agent state without locking.
+// The caller is responsible for locking.
+func (c *Agent) agentState() api.AgentState {
 	return c.session.AgentState
 }
 
@@ -290,6 +315,7 @@ func (c *Agent) Run(ctx context.Context, initialQuery string) error {
 					log.Info("Received input from channel", "userInput", userInput)
 					if userInput == io.EOF {
 						log.Info("Agent loop done, EOF received")
+						c.setAgentState(api.AgentStateExited)
 						return
 					}
 					query, ok := userInput.(*api.UserInputResponse)
@@ -337,6 +363,7 @@ func (c *Agent) Run(ctx context.Context, initialQuery string) error {
 				case userInput = <-c.Input:
 					if userInput == io.EOF {
 						log.Info("Agent loop done, EOF received")
+						c.setAgentState(api.AgentStateExited)
 						return
 					}
 					choiceResponse, ok := userInput.(*api.UserChoiceResponse)
@@ -594,7 +621,9 @@ func (c *Agent) Run(ctx context.Context, initialQuery string) error {
 func (c *Agent) handleMetaQuery(ctx context.Context, query string) (answer string, handled bool, err error) {
 	switch query {
 	case "clear", "reset":
+		c.sessionMu.Lock()
 		c.session.Messages = []*api.Message{}
+		c.sessionMu.Unlock()
 		return "Cleared the conversation.", true, nil
 	case "model":
 		return "Current model is `" + c.Model + "`", true, nil
--- a/pkg/ui/tui.go
+++ b/pkg/ui/tui.go
@@ -247,8 +247,7 @@ func (m model) Update(msg tea.Msg) (tea.Model, tea.Cmd) {
 		m.viewport.GotoBottom()
 	case tea.KeyMsg:
 		switch msg.Type {
-		case tea.KeyCtrlC, tea.KeyEsc:
-			fmt.Println(m.textarea.Value())
+		case tea.KeyCtrlC, tea.KeyEsc, tea.KeyCtrlD:
 			return m, tea.Quit
 		case tea.KeyEnter:
 			if m.agent.Session().AgentState == api.AgentStateWaitingForInput {
--- a/pkg/agent/mcp_client.go
+++ b/pkg/agent/mcp_client.go
@@ -72,6 +72,8 @@ func (a *Agent) UpdateMCPStatus(ctx context.Context, mcpClientEnabled bool) erro
 	}
 
 	// Update the session with MCP status
+	a.sessionMu.Lock()
+	defer a.sessionMu.Unlock()
 	a.session.MCPStatus = status
 
 	return nil
@@ -141,6 +143,8 @@ func (a *Agent) convertMCPStatus(mcpStatus *mcp.MCPStatus) *api.MCPStatus {
 // GetMCPStatusText returns a formatted text representation of the MCP status
 // This can be used by UIs that want to display the status as text
 func (a *Agent) GetMCPStatusText() string {
+	a.sessionMu.Lock()
+	defer a.sessionMu.Unlock()
 	if a.session.MCPStatus == nil {
 		return ""
 	}
--- a/pkg/tools/bash_tool.go
+++ b/pkg/tools/bash_tool.go
@@ -126,7 +126,7 @@ func (t *BashTool) Run(ctx context.Context, args map[string]any) (any, error) {
 		cmd.Env = append(cmd.Env, "KUBECONFIG="+kubeconfig)
 	}
 
-	return executeCommand(cmd)
+	return executeCommand(ctx, cmd)
 }
 
 type ExecResult struct {
@@ -163,7 +163,7 @@ func IsInteractiveCommand(command string) (bool, error) {
 	return false, nil
 }
 
-func executeCommand(cmd *exec.Cmd) (*ExecResult, error) {
+func executeCommand(ctx context.Context, cmd *exec.Cmd) (*ExecResult, error) {
 	command := strings.Join(cmd.Args, " ")
 
 	if isInteractive, err := IsInteractiveCommand(command); isInteractive {
@@ -177,7 +177,7 @@ func executeCommand(cmd *exec.Cmd) (*ExecResult, error) {
 	// Handle streaming commands
 	if isWatch || isLogs || isAttach {
 		// Create a context with timeout
-		ctx, cancel := context.WithTimeout(context.Background(), 7*time.Second)
+		timeoutCtx, cancel := context.WithTimeout(ctx, 7*time.Second)
 		defer cancel()
 
 		// Create pipes for stdout and stderr
@@ -230,7 +230,7 @@ func executeCommand(cmd *exec.Cmd) (*ExecResult, error) {
 
 		// Wait for either timeout or command completion
 		select {
-		case <-ctx.Done():
+		case <-timeoutCtx.Done():
 			isTimeout = true
 			// Kill the process immediately on timeout
 			if cmd.Process != nil {
--- a/pkg/tools/custom_tool.go
+++ b/pkg/tools/custom_tool.go
@@ -142,7 +142,7 @@ func (t *CustomTool) Run(ctx context.Context, args map[string]any) (any, error) 
 	cmd.Dir = workDir
 	cmd.Env = os.Environ()
 
-	return executeCommand(cmd)
+	return executeCommand(ctx, cmd)
 }
 
 // CheckModifiesResource determines if the command modifies resources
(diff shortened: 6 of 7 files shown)


cmd/main.go
```diff
@@ -524,12 +524,74 @@ type session struct {
...
+		return fmt.Errorf("running UI: %w", err)
+	}
+
+	err = s.agent.Run(ctx)
+	if err != nil {
+		return fmt.Errorf("running conversation: %w", err)
+	}
+
+	<-ctx.Done()
+	return nil
```
justinsb: Nit: maybe return ctx.Err() ?  But TBH I don't know what the "right" thing is here, because the error came from something we were passed so /shrug

cmd/main.go
```diff
@@ -616,16 +678,16 @@ func (s *session) answerQuery(ctx context.Context, query string) error {
 		s.doc.AddBlock(infoBlock)
 
 	case query == "tools":
-		if s.conversation == nil {
+		if s.agent == nil {
 			return fmt.Errorf("listing tols: conversation is not initialized")
```
justinsb: Nit: tools (pre-existing typo)

pkg/agent/conversation.go
```diff
@@ -73,9 +100,24 @@ type Conversation struct {
 	workDir string
 }
 
-func (s *Conversation) Init(ctx context.Context, doc *ui.Document) error {
+func (s *Agent) Init(ctx context.Context, doc *ui.Document) error {
 	log := klog.FromContext(ctx)
 
+	s.Input = make(chan any, 10)
+	// TODO: this need to be probably a buffered channel because
```
justinsb: Still a TODO?  Looks like you did the buffering.  And I agree with buffering :-)

pkg/agent/conversation.go
```diff
@@ -135,8 +177,419 @@ func (c *Conversation) Close() error {
 	return nil
 }
 
+func (c *Agent) Run(ctx context.Context) error {
+	log := klog.FromContext(ctx)
+
+	log.Info("Starting agent loop")
+
+	if c.currIteration == 0 && c.state == AgentStateIdle {
+		c.Output <- ui.NewAgentTextBlock().WithText("Hey there, what can I help you with today?")
```
justinsb: Oooh clever!

pkg/agent/conversation.go
```diff
@@ -135,8 +177,419 @@ func (c *Conversation) Close() error {
...
+
+			if userInput == io.EOF {
+				log.Info("Agent loop done")
+				return
+			}
+
+			if userInput != nil {
+				log.Info("User input", "userInput", userInput)
+				switch u := userInput.(type) {
+				case string:
```
justinsb: Nit: can we make these strongly typed?  Does a type-alias suffice, or do we need a full struct?

pkg/agent/conversation.go
```diff
@@ -135,8 +177,419 @@ func (c *Conversation) Close() error {
...
+							if err != nil {
+								log.Error(err, "error executing action", "output", output)
+								c.state = AgentStateDone
+								c.pendingFunctionCalls = []ToolCallAnalysis{}
+								continue
+							}
+
+							// Handle timeout message using UI blocks
+							if execResult, ok := output.(*tools.ExecResult); ok && execResult != nil && execResult.StreamType == "timeout" {
+								infoBlock := ui.NewAgentTextBlock().WithText("\nTimeout reached after 7 seconds\n")
```
justinsb: Nit: avoid hard-coding 7 in the message?  (Maybe just "Function call timed out"?)

pkg/agent/conversation.go
```diff
@@ -135,8 +177,419 @@ func (c *Conversation) Close() error {
...
+				}
+
+				// Clear our "response" now that we sent the last response
+				c.currChatContent = nil
+
+				if c.EnableToolUseShim {
+					// convert the candidate response into a gollm.ChatResponse
+					stream, err = candidateToShimCandidate(stream)
+					if err != nil {
+						c.state = AgentStateDone
```
justinsb: Given goroutines are basically free, I wonder if it's easier to "unroll" the state machine.

pkg/agent/conversation.go
```diff
@@ -135,8 +177,419 @@ func (c *Conversation) Close() error {
...
+					// remainining part of the loop will be now resumed when we receive a choice input
+					// from the user.
+					c.state = AgentStateWaiting
+					continue
+				}
+
+				// now we are in the clear to dispatch the tool calls
+
+				// execute all pending function calls
+				for _, call := range c.pendingFunctionCalls {
```
justinsb: It looks like this code is duplicated and I've lost track of why...

pkg/agent/conversation.go
```diff
@@ -135,8 +177,419 @@ func (c *Conversation) Close() error {
...
+				modifiesResourceStr = llmModifies
+			}
+		}
+		toolCallAnalysis[i].ModifiesResourceStr = modifiesResourceStr
+		toolCallAnalysis[i].ParsedToolCall = toolCall
+	}
+	return toolCallAnalysis, nil
+}
+
+func (c *Agent) handleChoice(ctx context.Context, choice int32) (dispatchToolCalls bool) {
```
justinsb: Nit: we really need an "enum" instead of int32

pkg/agent/conversation.go
```diff
@@ -135,8 +177,419 @@ func (c *Conversation) Close() error {
...
+		c.Output <- ui.NewErrorBlock().SetText("Invalid choice received. Cancelling operation.")
+		c.pendingFunctionCalls = []ToolCallAnalysis{}
+		dispatchToolCalls = false
+	}
+	return dispatchToolCalls
+}
+
 // RunOneRound executes a chat-based agentic loop with the LLM using function calling.
-func (a *Conversation) RunOneRound(ctx context.Context, query string) error {
+func (a *Agent) RunOneRound(ctx context.Context, query string) error {
```
justinsb: Do we need RunOneRound and Run ?

droot:
@zvdy @tuannvm @hakman @noahlwest This is a major refactor, would be great if you guys can share feedback (review as well try out the html/tui interfaces).

@tuannvm I have a TODO to test the mcp-client support and add the status message.

noahlwest:
> @zvdy @tuannvm @hakman @noahlwest This is a major refactor, would be great if you guys can share feedback (review as well try out the html/tui interfaces).
> 
> @tuannvm I have a TODO to test the mcp-client support and add the status message.

Ack. Will take a look today.

noahlwest:
I'll take a look at the code later. Overall testing this out, the changes seem great. The web interface in particular is pretty awesome.

For now; some initial feedbacks on the UX:
(Maybe important: this is in bash on debian-based distro)

nit picking: While assistant is trying to execute something, if you type it keeps redrawing the in-progress spinner. Not sure if it's redrawing the entire page or just that piece.

I was able to get the web interface stuck by getting a command to run that asks for sudo password (asked it to install helm). After entering password it continued on perfectly fine. It just didn't have an indication that was waiting on anything.

nit picking: When system is thinking, the web version puts the spinner a bit under screen. - 
image

yamls are a bit hard to parse in web interface - 
image


bubbletea keeps inserting `]11;rgb:1717/1414/2121` into my prompt, once on startup, then many times after that. Looks like it's trying to add some coloring either to my prompt or the terminal. - 
image

bubbletea does not handle resizing terminal very well it seems. It will show partial output, scrolling up and down forced it to refresh and fixed it. edit: this seems to work well on mac. This may just be a me problem, would like to know if anyone else runs into this.

bubbletea exits when you press q, even if you're just trying to add the letter q to a prompt. Also, no indicator of how to exit bubbletea when users are done.

sometimes when I enter a prompt with bubbletea, it just pauses for some seconds (maybe 15) with no feedback. Scrolling down sometimes forces it to update, but also adds a lot of those rgb:1717... into the prompt space. Not sure how to recreate this consistently


I can create issues for these later today OR I can wait till this is merged and see what is still there after.

pkg/agent/conversation.go
```diff
@@ -119,275 +193,637 @@ func (s *Conversation) Init(ctx context.Context, doc *ui.Document) error {
...
+				case *api.UserInputResponse:
+					log.Info("Text input", "text", query.Query)
+
+					c.sendMessage(api.MessageSourceUser, api.MessageTypeText, query.Query)
+					if c.session.AgentState == api.AgentStateIdle || c.session.AgentState == api.AgentStateDone {
+						log.Info("Transitioning to running state", "fromState", c.session.AgentState, "input", query.Query)
+
+						// we don't need the agentic loop for meta queries
+						// for ex. model, tools, etc.
+						answer, err := c.handleMetaQuery(ctx, query.Query)
```
justinsb: Nit: It might be more fluent to return (string, bool, error) where the bool is true iff the query was handled

droot:
@noahlwest Thank you so much for trying it out and sharing the findings. very helpful. I will try to tackle the critical ones tomorrow and file issues for the un-addressed ones. The good thing is both `tui` and `web` are optional interfaces.

zvdy:
Not a UI user myself but one thing I'm missing is darkmode, I implemented it  _(button on the top right)_

image

Will open a PR for it on your fork directly so you can pull from it:
`https://github.com/zvdy/kubectl-ai/tree/agent-native-ux`

and maybe decoupling `css and js` from the `index.html` although its simple enough to keep it in a single file, up to you

## [bubbletea](https://github.com/charmbracelet/bubbletea) issues
in my case the insertion for the `tui` its constant, using `zsh` related to [this issue](https://github.com/charmbracelet/bubbletea/issues/797) probably 

image

droot:
Thanks @zvdy for trying out and feedback. I did more cleanup today to simplify the code. We can accept your dark-mode change in a follow up and I am leaving the UI fixes for tui/web out for now because these are optional interfaces anyways.

mikebz:
Excited for this to merge!

tuannvm:
Thanks @droot for the major refactor! I'm pretty busy this week, so I haven't had time to review it yet. I will take a close look at the MCP feature over the weekend in case this one gets merged.

droot:
@justinsb @noahlwest In the latest commit, I also updated the manifests/scripts for deploying to kind cluster. PTAL

cmd/main.go
```diff
@@ -439,45 +364,51 @@ func RunRootCommand(ctx context.Context, opt Options, args []string) error {
...
-		} else if err != nil {
-			klog.Warningf("Failed to retrieve MCP server status for REPL startup: %v", err)
+	var userInterface ui.UI
+	switch opt.UIType {
+	case ui.UITypeTerminal:
+		// since stdin is already consumed, we use TTY for taking input from user
+		useTTYForInput := hasInputData
+
+		var u ui.UI
+		u, err = ui.NewTerminalUI(k8sAgent, useTTYForInput, recorder)
```
noahlwest: nit: Any reason not to just `userInterface, err = ... ` instead of having `userInterface = u` a few lines lower?

dev/tasks/build-images
```diff
@@ -20,8 +20,12 @@ fi
 if [[ -z "${BUILDX_ARGS:-}" ]]; then
   BUILDX_ARGS="--load"
 fi
+if [[ -z "${DOCKER:-}" ]]; then
+  DOCKER="docker"
+fi
+echo "Using container tool: ${DOCKER}"
 
-docker buildx build ${BUILDX_ARGS} \
+${DOCKER} buildx build ${BUILDX_ARGS} \
```
noahlwest: Do docker alternatives accept this same syntax?

pkg/agent/conversation.go
```diff
@@ -119,275 +208,547 @@ func (s *Conversation) Init(ctx context.Context, doc *ui.Document) error {
...
+}
 
-			// Add the tool call result to maintain conversation flow
-			if a.EnableToolUseShim {
-				// If shim is enabled, format the result as a text observation
-				observation := fmt.Sprintf("Result of running %q:\n%v", call.Name, output)
-				currChatContent = append(currChatContent, observation)
-			} else {
-				functionCallRequestBlock.SetResult(output)
+func (c *Agent) handleMetaQuery(ctx context.Context, query string) (answer string, err error) {
```
noahlwest: Why do we handle these here, and `version` in BuildRootCommand() in cmd/main.go ? Should we add version here or is there a reason it's separate?

pkg/ui/tui.go
```diff
@@ -0,0 +1,384 @@
...
+	username string // cached username
+}
+
+func newModel(agent *agent.Agent) model {
+	ta := textarea.New()
+	ta.Placeholder = "Send a message..."
+	ta.Focus()
+
+	ta.Prompt = "┃ "
+	ta.CharLimit = 280
```
noahlwest, line 171: Does this mean we can only have prompts up to 280 chars in tui? Thinking of cases similar to setup-dev-cluster eval where the prompt is quite large.

justinsb:
A few nits but I vote merge & iterate!

cmd/main.go
droot: I don't think there is any reason, will do this. I also realized that I can move the `webUI.RunServer` part in the Run() method for the webUI and that will make the code to setup UI much simpler

dev/tasks/build-images
droot: I am testing this script with `podman` because gmac doesn't have docker. working well with podman so far.

pkg/ui/tui.go
droot: Ack. Will tackle these in a follow up.

pkg/agent/conversation.go
droot: This is fixed already, comment seems to be on the old code.

cmd/main.go
droot: this has been refactored :), so comment is outdated now.

cmd/main.go
droot: this has been refactored :), so comment is outdated now.

pkg/agent/conversation.go
droot: This is very subtle and was a bit mind bending for me because now we are giving agent true agency. This is the beginning of proactive/ambient agents :)

pkg/agent/conversation.go
droot: Yes, this is already done :)

pkg/agent/conversation.go
droot: This is also refactored :) so newer code has much less duplication.

pkg/agent/conversation.go
droot: Moved to choice response type, easier to evolve now.

pkg/agent/conversation.go
droot: This is already addressed. New code is much cleaner.

pkg/agent/conversation.go
droot: Ack. will address this.

droot:
> A few nits but I vote merge & iterate!

Thanks @justinsb Actually, I am realizing the code has evolved (and simplified) significantly since you reviewed it. I will address a few pointed out (ui variables and handleMetaQuery) and get this in. You can take another look next week on merged code, and we can address remaining ones.

justinsb:
>  You can take another look next week on merged code

That SGTM - thank you!

pkg/agent/conversation.go
```diff
@@ -65,17 +94,77 @@ type Conversation struct {
...
+// addMessage creates a new message, adds it to the session, and sends it to the output channel
+func (c *Agent) addMessage(source api.MessageSource, messageType api.MessageType, payload any) *api.Message {
+	message := &api.Message{
+		ID:        uuid.New().String(),
+		Source:    source,
+		Type:      messageType,
+		Payload:   payload,
+		Timestamp: time.Now(),
+	}
+	c.session.Messages = append(c.session.Messages, message)
```
tuannvm, line 139: Multiple goroutines can modify `session.Messages` slice simultaneously without synchronization. We should add mutex for session protection.

pkg/agent/conversation.go
```diff
@@ -119,275 +208,547 @@ func (s *Conversation) Init(ctx context.Context, doc *ui.Document) error {
...
+				case <-ctx.Done():
+					log.Info("Agent loop done")
+					return
+				case userInput = <-c.Input:
+					log.Info("Received input from channel", "userInput", userInput)
+					if userInput == io.EOF {
+						log.Info("Agent loop done, EOF received")
+						return
+					}
+					query, ok := userInput.(*api.UserInputResponse)
```
tuannvm, line 321: Should we have more scrutiny in input validation?

pkg/agent/conversation.go
droot: Good catch, done!

pkg/agent/conversation.go
droot: We should probably but not clear how and what to validate. we need to sanitize all inputs to the agent. user input is still safe, the output from the tool calls is a major case for attack vector.

cmd/main.go
droot: Done.

pkg/agent/conversation.go
droot: version can be added here too.. we will have to propagate to the agent to make it work. I wasn't sure if that is useful as meta query, so left it alone. It is useful as `./kubectl-ai version` command though.

PR #388, opened 2025-07-04
gollm: experiment with supporting richer "LLM URIs"

justinsb:
Only for ollama initially.

### DIFF
--- a/gollm/ollama.go
+++ b/gollm/ollama.go
@@ -41,7 +41,8 @@ const (
 )
 
 type OllamaClient struct {
-	client *api.Client
+	client       *api.Client
+	defaultModel string
 }
 
 type OllamaChat struct {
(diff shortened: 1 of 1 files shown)


gollm/ollama.go
```diff
@@ -58,10 +59,20 @@ var _ Client = &OllamaClient{}
...
 
+	ollamaURL := envconfig.Host()
+	if opts.URL != nil && opts.URL.Host != "" {
+		ollamaURL.Host = opts.URL.Host
+	}
+	client := api.NewClient(ollamaURL, httpClient)
+
+	defaultModel := defaultOllamaModel
+	if opts.URL != nil {
+		defaultModel = opts.URL.Query().Get("model")
```
droot, line 71: hmm...we should only override the defaultModel if the url has `model` query parameter.. .right ? or am I missing something.

PR #390, opened 2025-07-05
Allow making width configurable for kubectl-ai's terminal UI

RaiAnandKr:
With the default fixed width of 80 (in the glamour library), I find a lot of my screen real estate getting wasted and a lot of scrolling needed when I run kubectl-ai on my terminal, so I wanted to tackle this.

There are few different paths to tackle this and honestly I am bit dicey about the path I took but here is my thinking:
1. I didn't want to figure the terminal width using one of the standard ways (`term.GetSize` or `COLUMNS` env var) since then the change would affect most of the folks by default.
2. I didn't want to add a new `--width` flag just for this.
3. I went with a totally custom env var approach so that only folks who really want this can use it.

I am open to thoughts and folks preferring some other path over this path, LMK.

### DIFF
--- a/pkg/ui/terminal.go
+++ b/pkg/ui/terminal.go
@@ -54,12 +54,34 @@ type TerminalUI struct {
 
 var _ UI = &TerminalUI{}
 
+func getCustomTerminalWidth() int {
+	// Check for user-configured width via environment variable
+	if widthStr := os.Getenv("KUBECTL_AI_TERM_WIDTH"); widthStr != "" {
+		if width, err := strconv.Atoi(widthStr); err == nil && width > 0 {
+			return width
+		}
+		klog.Warningf("Invalid KUBECTL_AI_TERM_WIDTH value %q, using default", widthStr)
+	}
+
+	// Return 0 to indicate no custom width should be set (use glamour's default)
+	return 0
+}
+
 func NewTerminalUI(doc *Document, journal journal.Recorder, useTTYForInput bool) (*TerminalUI, error) {
-	mdRenderer, err := glamour.NewTermRenderer(
+	width := getCustomTerminalWidth()
+
+	options := []glamour.TermRendererOption{
 		glamour.WithAutoStyle(),
 		glamour.WithPreservedNewLines(),
 		glamour.WithEmoji(),
-	)
+	}
+
+	// Only add WordWrap if a valid width is configured
+	if width > 0 {
+		options = append(options, glamour.WithWordWrap(width))
+	}
+
+	mdRenderer, err := glamour.NewTermRenderer(options...)
 	if err != nil {
 		return nil, fmt.Errorf("error initializing the markdown renderer: %w", err)
 	}


droot:
Thanks @RaiAnandKr 

Interesting options :). Will think and share my thoughts.

In the meantime, we are adding support for `tui` (rich terminal UI) [PR](https://github.com/GoogleCloudPlatform/kubectl-ai/pull/384/files) and it is looking very promising and will help us support all rich UI along with the current simple terminal interface. If you are curious how it handles the column width, give it a try and let me know:

PR: https://github.com/GoogleCloudPlatform/kubectl-ai/pull/384/files


```sh
make build
./bin/kubectl-ai --user-interface tui

```

RaiAnandKr:
Hey @droot thank you for taking a look. Will try the tui. While we are talking about UIs, I wanted to seek your opinion on something related. So one of our production use case involves invoking kubectl-ai CLI programmatically through code to debug a k8s issue and scraping its stdout. For this case, I wish there was a `plain` UI (it might not look too different from the current terminal UI but at least we would be saved by the ANSI characters and what not which comes from the terminal render).
Now, I fully understand that kubectl-ai is primarily built for interactive k8s management (including debugging) in terminal (and other UIs) so my requirement might sound left field. But then I also think that in production environments (at least at my place), only a handful of people (the curious k8s savy folks) will run kubectl-ai in terminal interactively and majority of them would want to see some automation being built on top of kubectl-ai (which is what I am tackling).

I guess this also express this philosophical debate I am having in my head - kubectl-ai being an AI assisted tool to "interactively manage/administrate" k8s clusters Vs an AI assisted tool to comprehensively debug/triage a given issue in my k8s cluster (to whatever extent is possible for the agent). 
The two aren't really exclusive but the latter requires more dedicated effort towards enabling the agent to debug independently and comprehensively as much as possible in one shot, I guess.
Curious to know your thoughts on the specific question as well as your general take on my philosophical debate🫠. Thank you!

droot:
Will share longer response later, but quick one:

> So one of our production use case involves invoking kubectl-ai CLI programmatically through code to debug a k8s issue and scraping its stdout. For this case, I wish there was a plain UI (it might not look too different from the current terminal UI but at least we would be saved by the ANSI characters and what not which comes from the terminal render).

Absolutely yes, this is one of the most important use-cases (for us as well), so regular plain terminal will always be supported. Infact, one of the key goals of the project is to deliver a CLI that is composable with other unix-style shell utilities so that it be can be dropped into lightweight automation workflows quickly.

We will support three UIs (plain terminal (default), rich terminal (tui), and http). In the linked PR, we are able to achieve this with a simple refactoring which will also enable saving/resuming these sessions (and possibly share with other users as well: think being able to resume an ongoing troubleshooting session that was started by someone else).


Wanted to drop an early screenshot of the web UI from that PR. You can simply run it with

```sh
make build
./bin/kubectl-ai --user-interface html
```

screenshot (below)

image

droot (approved):

pkg/ui/terminal.go
```diff
@@ -54,12 +54,34 @@ type TerminalUI struct {
 
 var _ UI = &TerminalUI{}
 
+func getCustomTerminalWidth() int {
+	// Check for user-configured width via environment variable
+	if widthStr := os.Getenv("KUBECTL_AI_TERM_WIDTH"); widthStr != "" {
```
droot, line 59: I wonder if we also allow user to set the `KUBECTL_AI_TERM_WIDTH=auto` to use the current terminal size automatically. This will give us a way to try it out to see if that will be sensible default.

droot: we can get this in, and support "auto" in the follow up, so merging this in.

droot (approved):

hakman:
Awesome change, thanks! 😄

Issue #393, opened 2025-07-07
Ensure `kubectl-ai` is aware of `kubectl rollouts history` command when analyzing read/write permissions requirement.


droot:
@zvdy I recently ran into this where `kubectl-ai` prompted me for approval while running `kubectl rollout history` command. Pl. take a look whenever you get a chance.

yankay:
`kubectl rollout status` is also readOnly :-)

zvdy:
Interesting edge case, status and history are read only, where the rest (undo restart pause resume) modify resources.

Will look into it, thanks for letting me know!

yankay:
> `kubectl rollout status` is also readOnly :-)

HI @zvdy 
I have created a PR to solve the issue. Would you please help review it?

PR #394, opened 2025-07-08
feat: Ensure accurate read/write classification for kubectl sub commands like `kubectl rollout history`

yankay:
Fix: #393

Previously, all rollout commands were incorrectly categorized as write operations,
resulting in unnecessary approval prompts for read-only subcommands such as history and status.

This PR introduces dedicated mappings for subcommands like `rollout` :
It can also be applied to commands like:
- auth can-i (read-only, checks permissions)
- auth reconcile (write, updates permission rules)

### DIFF
--- a/pkg/tools/kubectl_filter.go
+++ b/pkg/tools/kubectl_filter.go
@@ -36,11 +36,27 @@ var (
 	writeOps = map[string]bool{
 		"create": true, "apply": true, "edit": true, "delete": true,
 		"patch": true, "replace": true, "scale": true, "autoscale": true,
-		"expose": true, "rollout": true, "run": true, "set": true,
+		"expose": true, "run": true, "set": true,
 		"label": true, "annotate": true, "taint": true, "drain": true,
 		"cordon": true, "uncordon": true, "debug": true, "attach": true,
 		"cp": true, "reconcile": true, "approve": true, "deny": true,
 		"certificate": true,
+	}
+
+	readOnlySubOps = map[string]map[string]bool{
+		"rollout": {
+			"history": true,
+			"status":  true,
+		},
+	}
+
+	writeSubOps = map[string]map[string]bool{
+		"rollout": {
+			"pause":   true,
+			"restart": true,
+			"resume":  true,
+			"undo":    true,
+		},
 	}
 )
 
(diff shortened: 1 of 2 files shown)


zvdy:
Looks good to me, perhaps add more tests, I assume you have tested manually, If not, please do so by inducing the LLM to these edge cases. Once done we can merge @droot

pkg/tools/kubectl_filter.go
```diff
@@ -145,21 +161,26 @@ func analyzeCall(call *syntax.CallExpr) string {
 	}
 
 	verb := args[verbPos]
+	subVerb := ""
+	if verbPos+1 < len(args) && !strings.HasPrefix(args[verbPos+1], "-") {
+		subVerb = args[verbPos+1]
```
droot: Not sure how likely it is but technically possible to specify the flags between `verb` and `subverb` :) `k --f1 rollout --f2 history ` and this check will miss it.

a few options:

1. we can extract verb and subverb in a single for loop (above) (we stop at verb today)
2. filter out flags from the args to create nonFlagArgs and then simply extract verb, subverb from the filtered args.

pkg/tools/kubectl_filter_test.go
```diff
@@ -47,6 +47,7 @@ func TestKubectlModifiesResource(t *testing.T) {
 			{"Logs with follow", "kubectl logs nginx -f", "no"},
 			{"Watch pods", "kubectl get pods --watch", "no"},
 			{"Watch pods short", "kubectl get pods -w", "no"},
+			{"Rollout status", "kubectl rollout status deployment nginx", "no"},
```
droot, line 50: would be good to add flags `k --f1 rollouts --f2 status --f3 deployment nginx`

yankay: Thanks @droot 

It has been changed :-)

I also noticed some edge cases, such as commands like `kubectl -n default get deployment` where the first non-flag argument isn't the verb.

However, I found that LLMs don't generate such commands, so this should be a rare case.
I'll leave this out of this PR for now.

yankay:
> Looks good to me, perhaps add more tests, I assume you have tested manually, If not, please do so by inducing the LLM to these edge cases. Once done we can merge @droot

I have tested it : 

1. The `rollout status`:
img_v3_02o1_38968372-c7b8-46c2-89ae-cbe930ae762g

2. The `rollout undo`
img_v3_02o1_bbe35e48-c678-44c6-b3e4-a57f450e37bg

pkg/tools/kubectl_filter.go
```diff
@@ -133,33 +149,39 @@ func analyzeCall(call *syntax.CallExpr) string {
...
 	klog.V(2).Infof("analyzeCall: found kubectl: %q", firstArg)
 
-	// Get the verb (first non-flag argument after kubectl)
-	verbPos := 1 // Start after kubectl at position 0
-	for verbPos < len(args) && strings.HasPrefix(args[verbPos], "-") {
-		verbPos++
+	// Find the verb: first non-flag argument after "kubectl"
+	verb := ""
+	subVerb := ""
+	for i := 1; i < len(args); i++ {
```
droot: nit: you can simply use range here because indices are no longer useful in this case :)

```Go

for _, arg := range args {

}

```

pkg/tools/kubectl_filter.go
```diff
@@ -133,33 +149,39 @@ func analyzeCall(call *syntax.CallExpr) string {
...
 	}
-
-	if verbPos >= len(args) {
+	if verb == "" {
 		klog.Warningf("analyzeCall: no verb found after kubectl in args: %v", args)
 		return "unknown"
 	}
 
-	verb := args[verbPos]
 	hasDryRun := hasDryRunFlag(strings.Join(args, " "))
```
droot: couldn't resist :)

we also can check for dry-run flags in the same loop all the way and can kill this helper function :)

pkg/tools/kubectl_filter_test.go
droot: Aah... good catch. I think this may bite us down the line. Do you mind filing an issue for us to track this so that we can look into it.

droot (approved):
Thanks @yankay for working on this and thanks @zvdy for the review.

I have minor comments. None of them are blockers so approving the PR. If you plan to address them, let me know otherwise I will go ahead with the merge in the morning tomorrow.

pkg/tools/kubectl_filter.go
yankay: Thanks @droot 

I merged the `verbs` and  `hasDryRun`  into a single loop and was planning to delete this function.

But I noticed this function has a test case with some good cases in it.
So I refactored the loop out and used this test to verify it. What do you think?

droot (approved):

PR #395, opened 2025-07-08
feat: modularize prompt resolution with new `promptFile`

prasad89:
This PR adds the ability to store task prompts in external files instead of embedding them directly in YAML configurations.

### Key Changes
- Added `PromptFile` field to `ScriptStep` struct
- Implemented `ResolvePrompt()` method to load prompts from files or inline content
- Refactored `setup-dev-cluster` task to use external `setup-dev-cluster.md` file
- Maintains full backward compatibility with existing inline prompts

### Benefits
- Cleaner YAML configurations
- Better maintainability for complex prompts
- Improved editing experience with markdown files
- Supports both absolute and relative file paths

### DIFF
--- a/k8s-bench/main.go
+++ b/k8s-bench/main.go
@@ -53,7 +53,35 @@ const (
 )
 
 type ScriptStep struct {
-	Prompt string `json:"prompt"`
+	Prompt     string `json:"prompt"`
+	PromptFile string `json:"promptFile"`
+}
+
+// ResolvePrompt resolves the prompt from either inline or file source
+func (s *ScriptStep) ResolvePrompt(baseDir string) (string, error) {
+	// If both are provided, promptFile takes precedence
+	if s.PromptFile != "" {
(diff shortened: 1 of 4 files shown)


k8s-bench/main.go
```diff
@@ -53,7 +53,35 @@ const (
...
 
 type ScriptStep struct {
-	Prompt string `json:"prompt"`
+	Prompt     string `json:"prompt"`
+	PromptFile string `json:"promptFile"`
+}
+
+// ResolvePrompt resolves the prompt from either inline or file source
+func (s *ScriptStep) ResolvePrompt(baseDir string) (string, error) {
+	// If both are provided, promptFile takes precedence
```
droot: I think we should simply fail when both are provided to avoid any confusion.

droot (changes requested):

k8s-bench/main.go
prasad89: resolved :)

Issue #392, opened 2025-07-07
eval: create-pod evals asks clarifying questions and fails to complete


droot:
We recently changed our system prompt to ask clarification questions while creating resources. I was running all `create-` evals today and found `create-pod` eval to be failing.

We should revisit and see if we should change the success criterion to be asking for clarifying questions.

/cc @zvdy @noahlwest

noahlwest:
I'm able to recreate this maybe 30% of the time running `create-pod` eval. 70% of the time it just creates the pod, 30% I get something like
```
I can create the web-server pod in the create-pod-test namespace, but I need
to                                                                          
know which image to use. Can I use the nginx:latest image?
```

One option would be to make the prompt more specific, though I think assuming `nginx:latest` is reasonable unless otherwise specified. This opens us up to be more granular if we wanted. Have this eval specifically for creating the requested pod, another eval that purposely leaves out the version where our goal is to get the clarifying question. This is probably the easiest option.

Another option would be to allow success to be from *either* verify script or expect.contains in task.yaml, and add some text we would expect to see in the clarifying question. This would require some logic change, and picking the right expected text might be a bit awkward. I've seen some very different phrasing in my few runs, sometimes it doesn't even ask it just tells you to give an image.

I've been thinking there's some room to improve how we handle eval runs, this issue points out a good example where we could improve. I'll spend some time soon to think about how we can improve.

zvdy:
@noahlwest great analysis, really appreciate it, poking around the outputs, works on my end, it's inconsistent, so I would focus on the evaluation refactoring unless it does not make sense for anyone.

I think the new system prompt works well on most cases at least with my LLMs and use cases, haven't seen any other issues other than this eval task, maybe adding more context to itself might be a solution, another option is to modify the system prompt and add a default way to treat and manage images, but that is a sketchy topic, some people will argue latest always and others might want specific ones in order for them not to break due to breaking changes in a specific image

zvdy:
Seems to be fixed, perhaps its closable 

```
Evaluation Results:
==================

Task: create-pod
  LLM Config: {ID:shim_disabled-openai-claude-3-5-sonnet-20241022 ProviderID:openai ModelID:claude-3-5-sonnet-20241022 EnableToolUseShim:false Quiet:true}
    success

Task: create-pod-resources-limits
[...]

PR #398, opened 2025-07-08
Change TaskPattern filter to use regexp

noahlwest:
Noticed when recreating #392 : running evals with `--task-pattern create-pod` will run create-pod, create-pod-resources-limits, and create-pod-mount-configmaps.

Now, it does the same thing, but we can also do stuff like `--task-pattern ^create-pod$` so it only runs create-pod eval.


droot (approved):
Thanks.

Issue #399, opened 2025-07-09
[Bug]: Handle edge case where first non-flag argument is not the verb (e.g., kubectl -n default get deployment)


yankay:
**Describe the bug**
Currently, the command parser assumes that the first non-flag argument in a kubectl command is the verb (e.g., "get", "describe"). However, there are cases such as `kubectl -n default get deployment` where the first non-flag argument is not the verb.

ref: https://github.com/GoogleCloudPlatform/kubectl-ai/pull/394#discussion_r2193867706

**To Reproduce**
It's difficult to reproduce because the LLM can automatically adjust the order. For example:
![Image](https://github.com/user-attachments/assets/848120b2-8e48-4089-9884-e7fae39d1b71)

And it can be reproduced by the test case, Add
```{"Create pod", "kubectl -n default create -f pod.yaml", "yes"}, ``` to https://github.com/GoogleCloudPlatform/kubectl-ai/blob/main/pkg/tools/kubectl_filter_test.go#L76
The test case will fail.

**Expected behavior**
The test case can run success with the new test.

**Additional context**
Add any other context or logs here.

**Comment by zvdy** [comment]
Interesting edge case, doesn't this default to `unknown` choice? This should be reflected in the logs. Will try to reproduce although as you mentioned this is ordering dependant so tricking the LLM into giving the expected response can get somewhat difficult. Anyhow, if I get some free time I will look into this, we need to check if many verbs appear on the for loop, or simply fallback to unknown and rely on LLM feedback if many verbs appear and add a few tests if possible

**Comment by yankay** [comment]
> Interesting edge case, doesn't this default to `unknown` choice? This should be reflected in the logs. Will try to reproduce although as you mentioned this is ordering dependant so tricking the LLM into giving the expected response can get somewhat difficult. Anyhow, if I get some free time I will look into this, we need to check if many verbs appear on the for loop, or simply fallback to unknown and rely on LLM feedback if many verbs appear and add a few tests if possible

You're right, the logs show:
```
I0709 06:02:51.626520  621628 kubectl_filter.go:171] analyzeCall: unknown op for verb="default" subVerb="deployment/nginx"
```
It was choiced as  `unknown`.

zvdy:
then that means the LLM is the one answering wrongly, system promt might need to be tweaked a bit, which LLM are you using? would you mind/can test others?

yankay:
[...]

PR #400, opened 2025-07-09
Bugfix/399 first non-flag argument parsing

zvdy:
# Summary

This PR addresses issue #399 by improving the detection of the verb in kubectl commands. Previously, the logic would incorrectly classify commands where flags appeared before the verb (e.g., `kubectl --context=prod get pods`). Now, if any flags are present before the verb, the command is marked as "unknown" to avoid misclassification and rely purely on LLM choice for general complex scenarios.

e.g:
- `kubectl --context=prod --namespace=app get pods`
- `kubectl --token=eyJhbGciOiJSUzI1NiIsImtpZCI6IiJ9... get pods", "unknown`
- `kubectl --namespace=default get pods`

## Key changes
- Updated `analyzeCall` to return "unknown" if any flags appear before the verb.
- Added and updated tests to cover this edge case.
- All tests pass, confirming the fix and improved coverage.


zvdy:
commit 7384c0a implements https://github.com/GoogleCloudPlatform/kubectl-ai/issues/399#issuecomment-3059881405 the system prompt, i would personally leave the `verbPos` for loop check just in case we can defer to unknown(llm choice) if somehow the flag appears betweeen `kubectl` and `verb` even though it shouldn't

I have tried many times, with many providers, many different prompts always same scenario

image

I decided to recover my qwen key, and test it as it seems to be more independant, and somehow it completely ignores the system prompt

image

cc: @yankay in case you want to test further with other qwen models, searched a bit on it but nothing much, just this hackernews [post](https://news.ycombinator.com/item?id=43828875) and this github [issue](https://github.com/ollama/ollama/issues/6873) which makes me thing the model itself is the root cause here 

I'm using dashscope, if that helps `export OPENAI_ENDPOINT=https://dashscope-intl.aliyuncs.com/compatible-mode/v1`

here is also a list  of models `https://www.alibabacloud.com/help/en/model-studio/what-is-qwen-llm`

As a suggestion, if you can and have the tokens, would encourage trying deepseek, maybe it works with system prompts better although cost might be higher..

@droot  regarding defaulting to user rather than llm you can simply replace the returns from `unknown` to `yes` (for sure there are cleaner solutions) but while testing it, it can get quite intrusive (but safer) 

image

I would merge if it looks good to you, as it fixes most of the llms kubectl missplacings, and perhaps work in a different issue the default behavior of unknowns and the filtering and see what we can defer to users/llms to have a balance between risks and UX. Also, discarded the kubectl cheatsheet point as adding more information to the system prompt seems like is not the problem, simply some of the models ignore it or don't behave as expected

pkg/agent/systemprompt_template_default.txt
```diff
@@ -39,6 +39,14 @@ If you have enough information to answer the query:
 - Decide on the next action: use a tool or provide a final answer.
 {{end}}
 
+## Command Structuring Guidelines:
+**IMPORTANT:**  
+- When generating kubectl commands, ALWAYS place the verb (e.g., get, apply, delete) immediately after `kubectl`.
```
droot, line 44: Since this is `kubectl` specific, this should go in the kubectl's tool description.

We should use generic tool guidelines here.

droot: I wonder if that will affect the performance.

pkg/tools/kubectl_filter.go
```diff
@@ -149,6 +149,19 @@ func analyzeCall(call *syntax.CallExpr) string {
 
 	klog.V(2).Infof("analyzeCall: found kubectl: %q", firstArg)
 
+	// Check for flags before the verb (first non-flag argument)
+	verbPos := -1
+	for i, arg := range args[1:] {
+		if !strings.HasPrefix(arg, "-") {
```
droot, line 154: It is safe to assume the verbs and subVerbs to not contain `=`

basically, if an arg starts with `-` and also contains `=`, we can skip.

droot (changes requested):
A few minor comments

pkg/agent/systemprompt_template_default.txt
zvdy: regarding performance, I assume yes, it will make the llm check the ordering, which computationally will consume resources, still, should not be concerning as this should be significant, at least as far as my tests have gone, feel free to give it a try against more complex tasks. But my assumption is that this is not something so complex to do for them to be a bottleneck

Maybe we can benchmark it with ollama, since I suppose there is where we will see more of a bottleneck depending on the user's GPU, API based llms don't really concern me as response times should be almost identical

This reminds me about a chat he had a while ago in another issue about compression and different provider's optimizations, and this [LLMLingua paper](https://arxiv.org/abs/2310.05736) which _`shows that the proposed approach yields state-of-the-art performance and allows for up to 20x compression with little performance loss.`_

(not saying we should implement this as it will take quite some time but worth having it in mind as a FR when the project is more mature)

zvdy: regarding tool description it seems to ignore it when set as description in the `kubectl_tool.go` file itself, this [discussion](https://community.openai.com/t/prompt-structure-put-detail-in-function-description-or-user-prompt/943912/3) also seems to be related to this

tool desc. is for input output, at least with openai based llms, where system prompt are instructions that it will follow 

here is the test

image
 and desc snippet
 
 ```go
 "command": {
					Type: gollm.TypeString,
					Description: `The complete kubectl command to execute. Prefer to use heredoc syntax for multi-line commands. Please include the kubectl prefix as well.

IMPORTANT:
kubectl Command Structuring Guidelines:
- When generating kubectl commands, ALWAYS place the verb (e.g., get, apply, delete) immediately after kubectl.  
- Do NOT place flags or options before the verb.  
- Example:  
  - ✅ Correct: "kubectl get pods --namespace=default"
  - ❌ Incorrect: "kubectl --namespace=default get pods"  
- This ensures commands are properly recognized and filtered by the system.

IMPORTANT: Do not use interactive commands. Instead:
- Use 'kubectl get -o yaml', 'kubectl patch', or 'kubectl apply' instead of 'kubectl edit'
- Use 'kubectl exec' with specific commands instead of 'kubectl exec -it'
- Use service types like NodePort or LoadBalancer instead of 'kubectl port-forward'

 ```

pkg/tools/kubectl_filter.go
zvdy: `safe to assume verbs and subverbs will not contain =` 

yes, that is why the current implementation checks only for `-`, due to the fact that `=` _(arg starts with `-` and also contains `=`)_ does not cover all cases here is an example 

`kubectl -n default get pods`

so verbs will never start with - or --, flags only will, and some flags will contain or not =, both fall under the `if prefix -` check

pkg/agent/systemprompt_template_default.txt
zvdy: tried 3 different tool descriptions, none work, if you find any that could substitute the `system-prompt` let me know and i will happily implement it !

pkg/tools/kubectl_filter.go
zvdy: will mark as resolved,

I added a test to cover also the flag before verb without equal, just in case

droot: My bad, I should have stated it better.

I was trying to avoid returning "unknown" for the case `kubectl --context=prod get pods`, `kubectl --namespace=web get deployment` where flag is the format of `has prefix ('-') and also contains '='`

Essentially there are three types of flags:
1. Boolean. `kubectl --enable-something`
2. Spaced (key-value) `kubectl -n default ..., kubectl --context prod ...`
3. Non-spaced (key-value) `kubectl --namespace=default ..., kubectl --context=prod ...`

I was suggesting to return "unknown" for 1) and 2).

Since we are strongly steering the LLMs in system prompt, I am ok, skipping this, but wanted to clarify what I meant. Let me know if you want to merge this as-is or want to address this (this may require updating system prompt as well).

pkg/agent/systemprompt_template_default.txt
droot: Thanks for trying these out. Lets go ahead with this system prompt.

pkg/tools/kubectl_filter.go
zvdy: Makes sense now, thanks for clarifying, still for the 3rd edge case I still have a scenario like:

```
kubectl --replicas=3 scale deployment/nginx-deployment
```

that modifies resources 

But exist no infra modifying cases too

```
kubectl --namespace=web get deployment
```

So I would need to parse flags somehow and classify between modifying and non modifying 

Similarly to what we do for verbs 

I will give this a try tomorrow, even though as models evolve, the system prompt should be more than enough to guardrail them(most of them today work, at least from my tests except qwen), having the check does not harm anyone :-)

droot: Thanks for sharing both the mutating/non-mutating examples. My suggestion is to use the logic of `hasPrefix('-') and contains('=')` to skip the flag to extract verb/sub-verb. whether a verb/subverb is mutating or not is still determined by the current book-keeping we have.

Overall, I agree, the changes in this PR only improves the agent, so I am happy with the progress. 

I also wonder, if we can feed this context (our discussion) to agentic coding tools and see what they come up with :)

zvdy: Thanks for the feedback, I've implemented the suggested logic 

- flags before the verb are now checked for both prefix `-` and contains `=`
- only non spaced key-value pair flags (with =) are allowed before the verb for extraction
- boolean and key-value flags will result in unknown

the mutating/non-mutating logic still relies on the current implementation _(verb based)_ so no need to classify flags there _(I was making it more complex with my previous implementation, for some reason I assumed flag checks and classification were needed)_

It _LGTM_, feel free to merge!

droot: Awesome! This is exactly what we needed. Thank you for working through it.

droot (approved):

PR #402, opened 2025-07-09
Change task expectations to use regexp

ShubyM:
In order to fix the flakiness in #392 we need to support checking against multiple outputs which can be accomplished through regex. This change also results in eval behavior changing when both verify and expect are specified, the result will be the logical OR between them.

### DIFF
--- a/k8s-bench/eval.go
+++ b/k8s-bench/eval.go
@@ -250,11 +250,9 @@ func evaluateTask(ctx context.Context, config EvalConfig, taskID string, task Ta
 		return result
 	}
 
-	expectationsMet := false
 	var expectationFailures []model.Failure
-	if len(task.Expect) > 0 {
-		expectationsMet = true
 
+	if len(task.Expect) > 0 {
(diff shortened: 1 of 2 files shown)


droot:
@noahlwest PTAL.

k8s-bench/eval.go
```diff
@@ -243,12 +243,58 @@ func evaluateTask(ctx context.Context, config EvalConfig, taskID string, task Ta
...
 	// Run the agent
-	if err := x.runAgent(ctx); err != nil {
+	agentOutput, err := x.runAgent(ctx)
+	if err != nil {
 		// Unexpected error
 		result.Error = err.Error()
 		return result
 	}
 
+	expectationsMet := false
```
noahlwest: We could remove `expectationsMet` and check `len(expectationsFailures) == 0 && len(task.Expect) > 0` at 296 instead to simplify.

noahlwest (approved):
1 small nitpick otherwise LGTM

Issue #264, opened 2025-05-21
Idea: Running kubectl-ai in a container


janetkuo:
Wrapping `kubectl-ai` in a container can be useful in certain scenarios, for example: 

1. Portability and dependencies encapsulation
2. Integration with CI/CD pipelines or IDPs
3. Self-healing/remediation: in-cluster controller diagnoses problems with workloads and fixes errors automatically 
4. Quickly gather information about the cluster for observability or when there's an incident, such as summarizing issues / events / logs with context (could be triggered by alerts)
5. Integrate `kubectl-ai` into internal chat tools (e.g. Slack) for ChatOps 

... and more

btwseeu78:
I think the core logic should be separated from the ui of choice, let the core logic runs as a MCP tools itself, which will make it easier to plugin with all sort of pre-made ui. terminal is fine ,but this dependency makes it harder to adopt also due to rbac restrictions people will get all sorts results as per their access

droot:
> I think the core logic should be separated from the ui of choice, let the core logic runs as a MCP tools itself, which will make it easier to plugin with all sort of pre-made ui. terminal is fine ,but this dependency makes it harder to adopt also due to rbac restrictions people will get all sorts results as per their access

@btwseeu78 we made progress on decoupling UI logic from the core, the PR was merged last week. xref: https://github.com/GoogleCloudPlatform/kubectl-ai/pull/384 so now it is possible to expose the kubectl-ai itself as an MCP tool.

@ShubyM we already have scripts under `dev`. Two things need to happen:

- have a github action to publish a given release to github container registry (its okay for now for us to manually press the button in github action to publish this after a release is done). [script](https://github.com/GoogleCloudPlatform/kubectl-ai/blob/main/dev/tasks/build-images). FYI: I have been using `podman` to test the container image.
- Add `gcloud` tool to the image as well.
- Add a section in our main README showing how to use the image in a workflow

btwseeu78:
@droot i have added the provision to support for SSE or STDIO, i thought it would be nice to expose via http endpoint. 

https://github.com/GoogleCloudPlatform/kubectl-ai/issues/419

PR #404, opened 2025-07-14
feat: ghcr publish workflow

ShubyM:
Github action to publish kubectl-ai image to GHCR, part of #264. 
### Testing locally using [act](https://github.com/nektos/act)

1. In the docker/build-push-action step, set push: false 
2. (Optional) Add load: true in the same step to load the built image into your local Docker daemon for inspection
3. Run the following command 

```bash 
act -j build-and-publish
```


droot:
@ShubyM can you update the description with the details on how you tested it locally ?

PR #412, opened 2025-07-16
[WIP] Feat/docker image with gcloud

ShubyM:
Adding gcloud to kubectl-ai image to support work in #264


droot:
what is the size of the image ?
If it is too big, can we do something to reduce the size ?

PR #418, opened 2025-07-17
Enabled SSE mode support for mcp server

btwseeu78:
The current implementation allows you use mcp in STDIO modes. 

- [x] add options to change mode to sse
- [x] add options to define port to expose sse endpoint

### DIFF
--- a/dev/tasks/format.sh
+++ b/dev/tasks/format.sh
@@ -1,4 +1,4 @@
- #!/usr/bin/env bash
+#!/usr/bin/env bash
 # Copyright 2025 Google LLC
 #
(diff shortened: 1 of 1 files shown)


btwseeu78:
Proposal:

https://github.com/GoogleCloudPlatform/kubectl-ai/issues/419

dev/tasks/format.sh
```diff
@@ -1,4 +1,4 @@
-#!/usr/bin/env bash
+ #!/usr/bin/env bash
```
yankay: Why is there a space here?

btwseeu78: typo,  fixed

droot (approved):

PR #423, opened 2025-07-18
feat: Update docker image to reduce size (1.7gb -> 685mb) + docs

ShubyM:
This PR updates the existing docker image to reduce size by half and gives an example of how to run it against a GKE cluster

### DIFF
--- a/README.md
+++ b/README.md
@@ -295,6 +295,42 @@ A custom tool definition for `helm` could look like the following example:
     Use `helm --help` or `helm <subcommand> --help` to see full syntax, available flags, and examples for each command.
 ```
 
+## Docker Quick Start 
+This project provides a Docker image that gives you a standalone environment for running kubectl-ai, including against a GKE cluster.
+
+### Running the container against GKE
+
+#### Step 1: Build the Image
+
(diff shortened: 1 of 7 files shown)


README.md
```diff
@@ -295,6 +295,42 @@ A custom tool definition for `helm` could look like the following example:
...
+```bash
+docker run --rm -it \
+  -p 8080:8080 \
+  -v ~/.kube:/root/.kube:ro \
+  -e GOOGLE_CLOUD_PROJECT="YOUR_GCP_PROJECT_ID" \
+  -e GOOGLE_CLOUD_LOCATION="YOUR_GCP_LOCATION" \
+  kubectl-ai:latest \
+  --llm-provider vertexai \
+  --ui-listen-address 0.0.0.0:8080 \
+  --ui-type web
```
droot, line 331: We should also add examples for:

1. Running with just the GEMINI_API_KEY
2. default interactive terminal mode.

It will make sense to add `docs/container.md` file where we document all these examples and link it out from the main README. The main readme should include the basic example for basic terminal UI.

Doing those in a follow up is totally fine, so I am getting this in.

PR #424, opened 2025-07-18
treat kubectl exec to be a write operation

droot:
executing "exec" command should be treated as write operation.


pkg/tools/kubectl_filter_test.go
```diff
@@ -67,10 +67,11 @@ func TestKubectlModifiesResource(t *testing.T) {
 			{"Annotate", "kubectl annotate pods nginx description='my nginx'", "yes"},
 			{"Rollout restart", "kubectl rollout restart deployment nginx", "yes"},
 			{"Set image", "kubectl set image deployment/nginx nginx=nginx:latest", "yes"},
+			{"Exec into pod", "kubectl exec -n demo tgi-gemma-deployment-6f4954f5b7-8kcl4 -- nvidia-smi", "yes"},
 			{"Taint node", "kubectl taint nodes node1 key=value:NoSchedule", "yes"},
 			{"Run pod", "kubectl run nginx --image=nginx", "yes"},
 			{"Config set-context", "kubectl config set-context my-context", "no"},
-			{"Exec command", "kubectl exec nginx -- ls", "unknown"},
+			{"Exec command", "kubectl exec nginx -- ls", "yes"},
```
zvdy: Precisely thought this test was the edge case, ls doesn't modify resources per-se that was why I wanted to treat as unknown, but with the chat we had a few issues ago, if we want to move towards a more infrastructure safe approach makes sense to change from unknown to yes, but, we could also simply change the ls to a `rm -rf /` or something similar just for clarity

zvdy (commented):
minor test change

zvdy:
lgtm upon test change,makes sense to be extra safe with exec, maybe in the future we could rely more in LLM and execute improve UX

The scenarios in mind for me were sending requests to services following some guidelines checking logs, maybe some log filtering and so on..

pkg/tools/kubectl_filter_test.go
droot: Yes. `Safe by default` is the motivation for this change. Changed to `rm -rf /` for clarity.

droot:
Thanks for the review @zvdy. 

I expect LLMs to get better and be more reliable (better than humans) in assessing the effects of a command execution, so the current filtering mechanism is temporary and useful in this transitioning window only (bitter lessons). Other thing that is making things hard for our project is since we try to support models of different capabilities, it is difficult to reason how well one can depend on the LLM to make this judgement.

I think this is an area where we can use RL to fine-tune a very small model to detect write or not and get very accurate and decouple from the main LLM.

PR #430, opened 2025-07-21
feat: surface error messages from vertex-ai-client

ShubyM:
Noticed while doing some of the containerization that some of the errors from the vertex-ai client would not be surfaced, such as credentials being missing:

Before: 
Screenshot 2025-07-21 at 3 01 37 PM

After:
Screenshot 2025-07-21 at 2 54 50 PM


droot (approved):
Thank you so much. Actually, I was about to ask you to help with this :)

PR #440, opened 2025-07-24
fix: make llm aware of mcp tools

ShubyM:
Previously, the agent registered function definitions with the LLM before initializing the MCP client. This resulted in only the built-in tools being passed to the model, leaving it unaware of any external tools available through MCP


droot (approved):
Good catch.

PR #444, opened 2025-07-28
fix: surface error from mcp tool calls

ShubyM:
Surface MCP tool call errors to the UI 

Before:
Screenshot 2025-07-28 at 3 32 32 PM

After: 
Screenshot 2025-07-28 at 3 37 34 PM

### DIFF
--- a/pkg/agent/conversation.go
+++ b/pkg/agent/conversation.go
@@ -377,6 +377,7 @@ func (c *Agent) Run(ctx context.Context, initialQuery string) error {
 							c.setAgentState(api.AgentStateDone)
 							c.pendingFunctionCalls = []ToolCallAnalysis{}
 							c.session.LastModified = time.Now()
+							c.addMessage(api.MessageSourceAgent, api.MessageTypeError, "Error: "+err.Error())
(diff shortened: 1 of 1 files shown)


pkg/agent/conversation.go
```diff
@@ -377,6 +377,7 @@ func (c *Agent) Run(ctx context.Context, initialQuery string) error {
 							c.setAgentState(api.AgentStateDone)
 							c.pendingFunctionCalls = []ToolCallAnalysis{}
 							c.session.LastModified = time.Now()
+							c.addMessage(api.MessageSourceAgent, api.MessageTypeError, "Error: "+err.Error())
```
droot, line 380: I see that we added the error in the dispatchToolCalls function, do we need to do it here as well ?

ShubyM: Adding it in dispatchToolCalls allows the UI to see that the tool call is done, meaning it won't continue to show a spinner with executing and adding it here sends the output to the UI in the next bubble. We don't need to send the same message in both places however

droot (approved):

Issue #448, opened 2025-07-31
[Feature]: terminal UI does not show output of the toolcall


droot:
Terminal UI shows that it is running a tool but does not display or give any hints about the output of the tool-call.

The agent uses the tool call output to get the next response from LLM. Most of the time it is not a problem because LLM's output in the next turn will be using the output from the previous tool-call, but in some instances we have seen that LLM assumes that user is also aware of the previous tool-call output and decides the omit the output/summary in the final response.

Now displaying tool-call output always will can get noisy and repetitive (because final response may also include it), so we need a better way to present the tool-call output.

droot:
xref: https://github.com/GoogleCloudPlatform/kubectl-ai/issues/445

PR #451, opened 2025-08-01
[WIP] feat: show tool call output in terminal ui

ShubyM:
Attempts to implement #448 by truncating the tool call output

Before: 
image

After:
image

### DIFF
--- a/pkg/ui/terminal.go
+++ b/pkg/ui/terminal.go
@@ -17,6 +17,7 @@ package ui
 import (
 	"bufio"
 	"context"
+	"encoding/json"
 	"errors"
 	"fmt"
 	"io"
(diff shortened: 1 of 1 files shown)


pkg/ui/terminal.go
```diff
@@ -420,3 +425,44 @@ func (u *TerminalUI) handleMessage(msg *api.Message) {
...
+			if outStr, ok := stdout.(string); ok {
+				return truncateString(outStr, 1000)
+			}
+		}
+		b, err := json.MarshalIndent(m, "", "  ")
+		if err == nil {
+			return truncateString(string(b), 1000)
+		}
+		return fmt.Sprint(m)
+	}
```
droot, line 452: curious why we have to check for string and map ouput, we can constraint the output to always follow `map` format.

pkg/ui/terminal.go
```diff
@@ -420,3 +425,44 @@ func (u *TerminalUI) handleMessage(msg *api.Message) {
 func (u *TerminalUI) ClearScreen() {
 	fmt.Print("\033[H\033[2J")
 }
+
+func formatToolCallResponse(payload any) string {
```
droot: nit: let it always return the non-truncated string, and caller can call truncateString on the ouput.

droot (commented):
I am wondering if we should add a flag to opt-in to display output (without truncation).

droot:
@ShubyM The change is looking good. 

While you are at it, can we also fix the alignment for tool request/response. Mainly following changes:

1. `Running: <tool>` should have a new line before it (no line spacing looks strange)
2. `Tool output` is not aligned with anything

I think if you render these through markdown you will get the line spacing and left/right spaceing automatically. We used to have that and I think it got lost during the refactoring.

ShubyM:
With markdown rendering on the tool call response we get an output that looks like this: 

image

I'm not sure if there is an easy way to fix the table alignment when using the markdown rendering

droot:
> With markdown rendering on the tool call response we get an output that looks like this:
> 
> image
> I'm not sure if there is an easy way to fix the table alignment when using the markdown rendering

This is an improvement over what we have, so merging it in. We might tweak the truncation limit to something lower.

Issue #439, opened 2025-07-24
[Bug]: exit doesn't exit


mikebz:
**Environment (please complete the following):**
- MacOS, zsh
-  current build 2025-07-24 of `kubectl-ai` built from source.

**Describe the bug**

When issuing an exit command the tool doesn't actually exit
```
>>> exit

  It has been a pleasure assisting you. Have a great day                      

>>> exit

  It has been a pleasure assisting you. Have a great day                      

>>> quit

  It has been a pleasure assisting you. Have a great day.                     



```

**To Reproduce**
1. Launch
2. Do something
3. Ask to exit.

**Expected behavior**
1. The agentic loop stops and I get back to the shell.

droot:
I was trying out latest version and discovered this is partially fixed.

The agent attached to the conversation is terminated, but the terminal UI doesn't exit. Re-opening it so that we can address it.

PR #452, opened 2025-08-03
feat: add Close method to UI interface and implement resource cleanup in REPL

prasad89:
In this PR I have addressed https://github.com/GoogleCloudPlatform/kubectl-ai/issues/439#issuecomment-3140773328 by adding Close method to UI interface and implement resource cleanup in REPL

Closes https://github.com/GoogleCloudPlatform/kubectl-ai/issues/439

### DIFF
--- a/pkg/ui/tui.go
+++ b/pkg/ui/tui.go
@@ -128,6 +128,13 @@ func (u *TUI) Run(ctx context.Context) error {
 func (u *TUI) ClearScreen() {
 }
 
+func (u *TUI) Close() error {
+	if u.program != nil {
+		u.program.Quit()
+	}
+	return nil
+}
+
 type resultMsg struct {
 	duration time.Duration
 	food     string
--- a/cmd/main.go
+++ b/cmd/main.go
@@ -468,6 +468,11 @@ func repl(ctx context.Context, initialQuery string, ui ui.UI, agent *agent.Agent
 		return fmt.Errorf("running UI: %w", err)
 	}
 
+	// Ensure UI resources are cleaned up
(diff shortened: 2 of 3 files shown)


pkg/ui/tui.go
```diff
@@ -128,6 +128,13 @@ func (u *TUI) Run(ctx context.Context) error {
 func (u *TUI) ClearScreen() {
 }
 
+func (u *TUI) Close() error {
```
noahlwest, line 131: (this may be out of scope of this pr, feel free to resolve or ignore if it is): Do we need a function like this to clean up web ui resources too? If we do, but it's out of scope for this pr, we can make an issue requesting it in a later pr.

noahlwest (changes requested):
Hey, I cloned this pr to manually test and I wasn't able to get it to work in a way that would close #439. To close 439, I would expect `exit` or `quit` to take me out of the terminal. I also tried with `--ui-type tui` since I see you have changes in tui, and still had the same non-closing behavior. I'll attach demos to this comment.

I made #453 before I saw this pr, though it closes the terminal on `exit` or `quit`, it does not clean up any resources like your PR does. Mike brought up a good point on that PR about adding a regression test since we've run into this problem more than once.

I think the changes in this PR are still good and help us follow best practices, I just didn't see the behavior I expected. If you could possibly demo the behavior that #439 expects, then feel free to resolve/ignore this.

terminal: https://github.com/user-attachments/assets/2a90f00a-ae50-4292-a453-7ad207950a86
tui: https://github.com/user-attachments/assets/820143c7-791a-4cc7-a69b-f651a190d224

PR #454, opened 2025-08-05
[WIP] feat: add mcp resource support

ShubyM:


droot:
This is awesome, will take a deeper look later.

I was curious if we can complete the operation by providing correct project-id and cluster name with billing stuff exported.

PR #456, opened 2025-08-07
feat: regression test for exit and quit

ShubyM:
This test fails on main but does work on this PR: #453

### DIFF
--- a/pkg/agent/conversation_test.go
+++ b/pkg/agent/conversation_test.go
@@ -1,0 +1,74 @@
+// Copyright 2025 Google LLC
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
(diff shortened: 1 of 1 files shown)


pkg/agent/conversation_test.go
```diff
@@ -0,0 +1,74 @@
...
+	for _, q := range tests {
+		t.Run(q, func(t *testing.T) {
+			ctx, cancel := context.WithCancel(context.Background())
+			defer cancel()
+
+			a := &Agent{
+				Input:   make(chan any, 1),
+				Output:  make(chan any, 10),
+				session: &api.Session{AgentState: api.AgentStateIdle},
+			}
```
droot, line 36: Something we should prototype --  if we can inject mocks for say tools and gollm, how far we can push the unit testing here. Feel free to spend sometime to explore this, because this can really improve the test coverage.

mikebz:
Thank you @ShubyM  for adding to the resiliency of our code base.

PR #470, opened 2025-08-12
[WIP] progres bar with context reminaing

ShubyM:


pkg/agent/conversation.go
```diff
@@ -546,6 +551,12 @@ func (c *Agent) Run(ctx context.Context, initialQuery string) error {
 				if streamedText != "" {
 					c.addMessage(api.MessageSourceModel, api.MessageTypeText, streamedText)
 				}
+
+				if pct, ok := gollm.ContextPercentRemaining(c.Model, usageMetadata); ok {
+					// c.addMessage(api.MessageSourceAgent, api.MessageTypeText, fmt.Sprintf("Context remaining: %.1f%%", pct))
+					c.addMessage(api.MessageSourceAgent, api.MessageTypeUsageMetadata, api.UsageMetadata{ContextPercentRemaining: pct})
```
droot: I think it will make more sense to just make this (contextRemaining) an attribute of the session object and update it whenever session's context is being updated. UI has access to the session object and we used that to communicate `AgentState` attribute.

pkg/api/models.go
```diff
@@ -52,6 +52,7 @@ const (
 	MessageTypeUserInputResponse  MessageType = "user-input-response"
 	MessageTypeUserChoiceRequest  MessageType = "user-choice-request"
 	MessageTypeUserChoiceResponse MessageType = "user-choice-response"
+	MessageTypeUsageMetadata      MessageType = "usage-metadata"
```
droot: We can avoid adding message type to communicate this. I am proposing we should add an attribute on the session.

pkg/api/models.go
```diff
@@ -24,6 +24,10 @@ type Session struct {
 	AgentState   AgentState
 	CreatedAt    time.Time
 	LastModified time.Time
+
+	// ContextPercentRemaining indicates the percentage of the model's context window that is still available.
+	ContextPercentRemaining float64
```
droot: Will it be better to store the actual consumed token and derive the percentage on demand. The reason is that the limits may change over time so tracking absolute unit will be more useful.

PR #471, opened 2025-08-12
eval: Add eval generator prompt

noahlwest:
Adds a prompt that can be used to generate eval files, given a task description.
* So far I've tried this gemini and chatgpt apps, and the gemini code assist vs code extension. It has worked pretty well with all of them.
* Task input section can be replaced as verbosely as you're willing to make it, and include any specifics you want. Example in #455 which used the following:
```
Prompt: "Our new checkout-service-green deployment in the e-commerce namespace has passed all tests. The current live version is checkout-service-blue. Can you switch all live traffic over to the green version now?"

Verification: The agent must identify the Service that routes traffic to the checkout application. It will find that the service selector is version: blue. The agent must patch the Service to change the selector to version: green. This will instantly redirect all traffic to the new deployment's pods.
```

I'm also thinking of adding an eval scaffolding tool to make the boilerplate a little easier.

Looking for feedback on any improvements that could be made to the prompt @droot @zvdy @prasad89 @ShubyM @justinsb @janetkuo

### DIFF
--- a/k8s-bench/eval-generator-prompt.md
+++ b/k8s-bench/eval-generator-prompt.md
@@ -1,0 +1,156 @@
+# ROLE AND GOAL
+
+You are an expert-level Site Reliability Engineer (SRE) with extensive Kubernetes experience and knowledge of best practices.
+
+Your goal is to generate a complete and self-contained evaluation task designed to test an AI agent's ability to perform Kubernetes-related tasks. You will produce a set of files that can be used to set up the scenario, present the problem to the AI, verify the solution, and clean up the environment.
+
+---
+
+# TASK CRITERIA
+
+Each generated evaluation must adhere to the following principles:
+
+1.  **Realistic:** The scenario must reflect a real-world problem an engineer would face. The prompt for the AI agent should be conversational and natural. Avoid unnecessary hints to help pass, and avoid providing info that lets the agent know it’s being tested.
+2.  **Self-Contained:** All Kubernetes resources MUST be created in a dedicated, unique namespace. This namespace should be defined as a variable (`NAMESPACE`) at the top of each shell script for consistency and to prevent conflicts.
+3.  **Verifiable:** The `verify.sh` script must contain specific, automatable checks (e.g., using `kubectl get ... -o jsonpath='...'` or `grep`). The script **must** `exit 0` on success and `exit 1` on failure.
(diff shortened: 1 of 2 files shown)


k8s-bench/eval-generator-prompt.md
```diff
@@ -0,0 +1,156 @@
...
+```
+
+
+
+**cleanup.sh**
+```bash
+#!/bin/bash
+set -e
+NAMESPACE={The exact same namespace as setup.sh}
+kubectl delete namespace $NAMESPACE --wait=false
```
janetkuo, line 63: This assumes the test is namespace-scoped

k8s-bench/eval-generator-prompt.md
```diff
@@ -0,0 +1,156 @@
...
+        image: busybox:1.36
+        # This command is invalid and will cause a crash
+        command: ["/bin/sh", "-c", "echo 'starting...' && sleep 5 && exit 1"]
+```
+
+
+**THE TASK**
+
+Now, using the role, criteria, format, and golden example above as your guide, generate a complete evaluation for the following user-provided task.\
+TASK: "{INSERT_EVALUATION_TOPIC_HERE}"
```
prasad89, line 154: Everything else looks good just a few questions:

1. Can we enhance the markdown further, considering that LLMs can understand it better?
2. Will evaluating only the topic be sufficient? Perhaps we could also include a description with do’s and don’ts, especially if we don’t plan to iterate further.

k8s-bench/eval-generator-prompt.md
```diff
@@ -0,0 +1,156 @@
...
+This is an example of a perfect output for the task: "Fix a Pod that is stuck in a CrashLoopBackOff state due to a bad command."
+**task.yaml**
+```yaml
+script:
+  - prompt: "Hey, I just deployed my 'finance-app' in the `finance-ns` namespace, but the pod seems to be stuck in a crash loop. Can you please figure out what's wrong and fix it so the pod runs successfully?
+setup: "setup.sh"
+verifier: "verify.sh"
+cleanup: "cleanup.sh"
+difficulty: "easy"
+```
```
zvdy, line 79: missing closing quote in prompt " 
Makes yaml possibly unreadable for small models

PR #472, opened 2025-08-12
eval: Add canary deployment eval

noahlwest:
Adds an eval for creating a canary deployment with some stated requirements.
* This eval fails 80% or more of the time on Gemini 2.5 pro, due to small mistakes in correctness. Wrong deployment names, incorrect number of replicas, scaling down the stable deployment even though explicitly told not to. The overall task is relatively simple but since these small measure of correctness are important, and even gemini 2.5 pro keeps making mistakes on them, I'm marking it as hard.

Generated using the prompt in #471 with this added input:
```
Prompt: "We want to test a new version of our recommendation engine (image tag v1.29) in production. Can you deploy it as a canary? We want about 10% of traffic to go to the new version. The current deployment is named engine-v2-0 and has 10 replicas."

Verification: The agent must create a new Deployment named engine-v2-1 with the new image tag and 1 replica. It must ensure the Service selector is updated to select pods from both deployments, for example, by using a common label like app: recommendation-engine that is present on both sets of pods, while version-specific labels differentiate them.
```

Example of a successful run's output:
```
deployment.apps/engine-v2-1 condition met
deployment.apps/engine-v2-1 condition met
deployment.apps/engine-v2-0 condition met
--> Waiting for service 'recommendation-engine' to have '11' endpoints...
Success: Service has all 11 endpoints.
--- Verification Successful! ---
Canary deployment is correctly configured and receiving traffic.
```

Example of a failed run's output:
```
deployment.apps/engine-v2-1 condition met
deployment.apps/engine-v2-1 condition met
error: timed out waiting for the condition on deployments/engine-v2-0
ERROR: Stable deployment does not have the correct amount of replicas.
```

### DIFF
--- a/k8s-bench/tasks/create-canary-deployment/cleanup.sh
+++ b/k8s-bench/tasks/create-canary-deployment/cleanup.sh
@@ -3,4 +3,4 @@ set -e
 NAMESPACE="canary-deployment-ns"
 
 # Delete the namespace
-kubectl delete namespace $NAMESPACE --wait=false
+kubectl delete namespace $NAMESPACE --wait=false --ignore-not-found
--- a/k8s-bench/tasks/create-canary-deployment/cleanup.sh
+++ b/k8s-bench/tasks/create-canary-deployment/cleanup.sh
@@ -1,0 +1,6 @@
+#!/bin/bash
+set -e
+NAMESPACE="canary-deployment-ns"
+
+# Delete the namespace
+kubectl delete namespace $NAMESPACE --wait=false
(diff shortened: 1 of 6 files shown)


k8s-bench/tasks/create-canary-deployment/cleanup.sh
```diff
@@ -0,0 +1,6 @@
+#!/bin/bash
+set -e
+NAMESPACE="canary-deployment-ns"
+
+# Delete the namespace
+kubectl delete namespace $NAMESPACE --wait=false
```
droot: Lets try to keep these scripts idempotent. For example, in this case, adding command line flag to ignore if the namespace doesn't exist.

We should add hint in our prompt because there might be other things related to idempotency.

k8s-bench/tasks/create-canary-deployment/task.yaml
```diff
@@ -0,0 +1,6 @@
+script:
+  - prompt: "We want to test a new version of our recommendation engine (image tag 1.29) in production without disturbing the existing stable deployment. Can you deploy it as a canary (as engine-v2-1)? We want about 10% of traffic to go to the new version. The current deployment is named engine-v2-0 and has 10 replicas."
```
droot: I am wondering about this extra information: `and has 10 replicas.`

droot (commented):
Looks good. I have some minor comments.

k8s-bench/tasks/create-canary-deployment/task.yaml
noahlwest: Removed this extra prompt hint.

k8s-bench/tasks/create-canary-deployment/cleanup.sh
noahlwest: Good point. Prasad made a good point about enhancing the prompt further since frontier models can handle it quite well, this would be a good addition.

droot (approved):

PR #475, opened 2025-08-12
feat: Remove goodbye message, make `sessions` metaquery use table format

noahlwest:
* Removes goodbye message since it seems out of place to close every interaction (including metaqueries)
* Makes `session` metaquery use table format instead of the previous stacked format. example:
image


droot (approved):
Thank you! Looks great.

PR #476, opened 2025-08-12
[WIP] feat: meta queries tests + mocks

ShubyM:


droot:
@ShubyM will get to this later today. Thank you.

pkg/agent/conversation.go
```diff
@@ -387,7 +387,17 @@ func (c *Agent) Run(ctx context.Context, initialQuery string) error {
...
+							c.addMessage(api.MessageSourceAgent, api.MessageTypeText, answer)
+						} else {
+							c.Output <- &api.Message{
+								ID:        uuid.New().String(),
+								Source:    api.MessageSourceAgent,
+								Type:      api.MessageTypeText,
+								Payload:   answer,
+								Timestamp: time.Now(),
+							}
+						}
```
droot, line 400: What exactly is happening here ? (addMessage vs direct entry to Output ?)

pkg/agent/conversation.go
```diff
@@ -471,6 +481,7 @@ func (c *Agent) Run(ctx context.Context, initialQuery string) error {
 					log.Error(err, "error sending streaming LLM response")
 					c.setAgentState(api.AgentStateDone)
 					c.pendingFunctionCalls = []ToolCallAnalysis{}
+					c.addMessage(api.MessageSourceAgent, api.MessageTypeError, "Error: "+err.Error())
```
droot, line 484: Aah... this is why the LLM errors were not showing up.

pkg/agent/agent_loop_test.go
```diff
@@ -0,0 +1,312 @@
...
+					receivedResponse = true
+				}
+			}
+		case <-timeout:
+			t.Fatal("timed out waiting for model response")
+		}
+	}
+
+	// Agent should be exited after one run
+	time.Sleep(100 * time.Millisecond)
```
droot: In the RunOnce case, the output channel is closed. So instead of sleep here, you can just range over the output channel. That is more deterministic test.

pkg/agent/agent_test_helpers.go
```diff
@@ -0,0 +1,143 @@
...
+	"context"
+
+	"github.com/GoogleCloudPlatform/kubectl-ai/gollm"
+	"github.com/GoogleCloudPlatform/kubectl-ai/pkg/api"
+	"github.com/GoogleCloudPlatform/kubectl-ai/pkg/sessions"
+	"github.com/GoogleCloudPlatform/kubectl-ai/pkg/tools"
+)
+
+// mockLLM is a mock implementation of the gollm.Client interface for testing.
+type mockLLM struct {
```
droot, line 27: I wonder if gomock can be a good option "go.uber.org/mock/gomock"

Something related https://github.com/GoogleCloudPlatform/kubectl-ai/pull/208

pkg/agent/conversation.go
ShubyM: Going directly to output prevents the message being added to chat history

droot: Can you say more about why we are doing this (why we don't want to add this to history especially that we added `clear` or `reset` to the history before handling the metaquery ?) if it helps, we can add an additional attribute `IsMeta` to help with filtering.

pkg/agent/conversation_test.go
```diff
@@ -0,0 +1,214 @@
...
+			defaultTools := tools.Default()
+			a := newTestAgent(&mockChat{}, (&defaultTools).AllTools()...)
+
+			if err := a.Run(ctx, ""); err != nil {
+				t.Fatalf("Run returned error: %v", err)
+			}
+
+			// Wait for the agent to prompt for input
+			timeout := time.After(time.Second)
+		WaitForPrompt:
```
droot, line 107: Pl. avoid using the go-to style labels, it makes it very hard to follow the code flow.

pkg/agent/conversation_test.go
```diff
@@ -0,0 +1,214 @@
...
+					if m.Type == api.MessageTypeUserInputRequest && m.Payload.(string) == ">>>" {
+						a.Input <- &api.UserInputResponse{Query: tc.query}
+						break WaitForPrompt
+					}
+				}
+			}
+
+			responseTimeout := time.After(time.Second)
+			var foundResponse bool
+		WaitForResponse:
```
droot, line 123: pl. avoid go-to labels.

pkg/agent/agent_loop_test.go
```diff
@@ -0,0 +1,313 @@
...
+					receivedResponse = true
+				}
+			}
+		case <-timeout:
+			t.Fatal("timed out waiting for error message")
+		}
+	}
+
+	for range a.Output {
+		if a.AgentState() != api.AgentStateExited {
```
droot, line 176: Should this be out of the for loop ? to ensure agent's final state to be in the exited ?

pkg/agent/agent_test_helpers.go
droot: Looking at the tests, I think gomock might help because the tests are making good use of injecting custom functions.

pkg/agent/agent_loop_test.go
```diff
@@ -0,0 +1,313 @@
...
+					receivedResponse = true
+				}
+			}
+		case <-timeout:
+			t.Fatal("timed out waiting for model response")
+		}
+	}
+
+	for range a.Output {
+		if a.AgentState() != api.AgentStateExited {
```
droot, line 136: I wonder if `runOnce` is true, should be test the terminal state check outside the for loop ?

pkg/agent/agent_loop_test.go
```diff
@@ -0,0 +1,313 @@
...
+			t.Errorf("Expected agent state to be Exited, but got %s", a.AgentState())
+		}
+	}
+}
+
+func TestAgentRun_LLMErrorResponse(t *testing.T) {
```
droot, line 142: I think TestAgentRun_LLMTextResponse and TestAgentRun_LLMErrorResponse can be combined as table driven form because the driver code is almost same.

pkg/agent/agent_test_helpers.go
droot: Also, I think it will be good to separate out the meta query tests from other tests in separate PRs, it makes it harder to review the PR.

ShubyM: Sounds good!

PR #478, opened 2025-08-13
feat: Added --show-tool-output to display tool's output

droot:
We recently added feature to show tool output (with truncation limit) by default. After a testing it out today, I am realizing that the terminal output is getting even more verbose and also causing duplication of information.

This PR restores the original behavior of not showing the tool output by default. Instead, now users can use `--show-tool-ouput` command line flag to always display the tool output. 

default behavior:

image

opt-in behavior:

image

/cc @noahlwest  @ShubyM 

Also adjusted the "Running" to align with other outputs.

### DIFF
--- a/cmd/main.go
+++ b/cmd/main.go
@@ -124,8 +124,8 @@ type Options struct {
 	ListSessions  bool   `json:"listSessions,omitempty"`
 	DeleteSession string `json:"deleteSession,omitempty"`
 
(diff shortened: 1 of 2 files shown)


ShubyM (approved):

noahlwest (approved):
Code looks good to me. For posterity it would be good to show the before/after

droot:
> Code looks good to me. For posterity it would be good to show the before/after

Good pt. Added screenshots.

PR #480, opened 2025-08-13
feat: Add resume-session to metaquery handling

noahlwest:
Adds "resume-session <session ID>" to metaquery handling.

demo (start from no session, load one session, load another session): https://github.com/user-attachments/assets/8ceca555-ab60-482a-8d81-e75074c437ab

demo (changing sessions in web ui): https://github.com/user-attachments/assets/49c4dd48-d13e-43e6-835f-c19d1421c9cf

### DIFF
--- a/pkg/agent/conversation.go
+++ b/pkg/agent/conversation.go
@@ -200,13 +200,7 @@ func (s *Agent) Init(ctx context.Context) error {
 	}
 
 	if session, ok := s.ChatMessageStore.(*sessions.Session); ok {
-		metadata, err := session.LoadMetadata()
-		if err != nil {
-			return fmt.Errorf("failed to load session metadata: %w", err)
-		}
-		s.session.ID = session.ID
-		s.session.CreatedAt = metadata.CreatedAt
-		s.session.LastModified = metadata.LastAccessed
+		s.loadSession(session.ID)
 	} else {
 		s.session.ID = uuid.New().String()
 		s.session.CreatedAt = time.Now()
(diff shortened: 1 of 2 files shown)


pkg/agent/conversation.go
```diff
@@ -737,6 +737,41 @@ func (c *Agent) handleMetaQuery(ctx context.Context, query string) (answer strin
...
+		metadata, err := session.LoadMetadata()
+		if err != nil {
+			c.sessionMu.Unlock()
+			return "", false, fmt.Errorf("failed to load session metadata: %w", err)
+		}
+		c.session.ID = session.ID
+		c.session.CreatedAt = metadata.CreatedAt
+		c.session.LastModified = metadata.LastAccessed
+		c.llmChat.Initialize(c.session.ChatMessageStore.ChatMessages())
+		c.sessionMu.Unlock()
```
droot: Functionality looks great. I think we can abstract out `loadSession(...)` functionality in some common place so that it can be called here and also in `agent.Init()`. what do you think ?

droot: I see there is some overlap with the `save-session` functionality, so if you want to combine this with the `save-session` PR, I am totally fine with it.

noahlwest: Added `loadSession()`, also removed an inaccessible code path. `save-session` had some other feedbacks that will make it a little bigger, so I'll keep them separate.

droot (approved):

PR #493, opened 2025-08-15
Fix Var reference for UI configuration for ui Type

iomarcovalente:
This fixes the reference var in the main README for ui config type.

To add to this: I found in code that there is a TUI type too, but as it is not working as it should (tested on Mac and for every input you get a weird string in the user input box) I haven't advertised it in the README.

### DIFF
--- a/README.md
+++ b/README.md
@@ -268,7 +268,7 @@ removeWorkdir: false             # Remove temporary working directory after exec
 kubeconfig: "~/.kube/config"      # Path to kubeconfig file
 
 # UI configuration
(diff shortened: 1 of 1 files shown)


droot:
> To add to this: I found in code that there is a TUI type too, but as it is not working as it should (tested on Mac and for every input you get a weird string in the user input box) I haven't advertised it in the README.

Yeah, tui has bunch of problems to be resolved for different terminal types etc. I hope someone from the community picks it up because it has ton of potential to provide great UX.

Issue #466, opened 2025-08-08
[Request]: Add tests for Gemini session persistence


noahlwest:
#443 added session persistence for Gemini, and we should have tests for it. Basic flows like saving and loading sessions, creating a new session, deleting a session, and listing sessions should be covered. Edge cases like invalid session ID and corrupted save file should also be covered.

nisranjan:
@noahlwest #497 closes this, let me know if there are still open points to this

PR #497, opened 2025-08-18
WIP for Session Tests

nisranjan:
@noahlwest 
Just implement Session Tests for #466 

It covers the following:

- Creating a Session
- Deleting a Session
- Listing Session
- Corrupted History file
- Concurrent Session Access
- Clear Session

A modification I made in session.go is to change the implementation of session file loading from entire JSON at one time to line by line so any corrupt JSON lines can be omitted 

Another issue that I found out during testing is that GetLatestSession gets the first session, however it gets the latest session when a call to ListSession has already been made which sorts the slice of sessions, at the moment I am not sure if there is a use case of calling GetLatestSession without calling the ListSession(), because if there is a case for that then we should fix this.


pkg/sessions/session_test.go
```diff
@@ -0,0 +1,376 @@
...
+		lastSession, err = manager.NewSession(Metadata{
+			ProviderID: "test-provider",
+			ModelID:    "test-model",
+		})
+		require.NoError(t, err)
+		time.Sleep(time.Millisecond) // Ensure different timestamps
+	}
+
+	// List sessions for correct order
+	manager.ListSessions()
```
noahlwest: Is this ListSessions call doing anything?

pkg/sessions/session_test.go
```diff
@@ -0,0 +1,376 @@
...
+
+// TestSessionPersistence tests the basic save and load functionality
+func TestSessionPersistence(t *testing.T) {
+	// Create a temporary directory for test sessions
+	tempDir, err := os.MkdirTemp("", "session-test-*")
+	require.NoError(t, err)
+	defer os.RemoveAll(tempDir)
+
+	// Create a session manager
+	manager := &SessionManager{BasePath: tempDir}
```
noahlwest, line 38: Looks like all these tests have some overlapping lines for setup, could we make this into a helper function and re-use it instead?

noahlwest (approved):
Thank you very much! Overall looks good, just two minor comments.

pkg/sessions/session_test.go
nisranjan: The problem as I mentioned earlier is that GetLatestSession() actually gets the First Session and not the Last or Latest Session. This is probably because of the way ListSession() sorts it and then GetLatestSession() returns the 0th item in slice / array. So this was causing assertion errors in the TestGetLatestSession().

I think that it would be better to add the session to a heap when Listing Sessions from the data store. I am curious to know what you think?

Also, I think you also need to decide weather I should remove this line of code now, in which case the tests will fail until u have fixed the issue. Or u can fix the issue and I will modify the tests after u have done that.

pkg/sessions/session_test.go
nisranjan: Great point, I have made changes in  [this commit] (https://github.com/GoogleCloudPlatform/kubectl-ai/pull/497/commits/a1c80acb9e4c89227ca16c9c0538d1cd815e648b). Take a look if this makes sense.

pkg/sessions/session_test.go
noahlwest: I made an issue for this and assigned myself, I'll fix this today. https://github.com/GoogleCloudPlatform/kubectl-ai/issues/518. I'll also leave a message here after the fix is submitted to keep you up to date.

pkg/sessions/session_test.go
noahlwest: Thanks, the changes there look good to me.

pkg/sessions/session_test.go
noahlwest: The fix for #518 is in, so things should be working as intended now.

nisranjan: I have updated the session_test.go and removed the unnecessary call of ListSession() in TestGetLatestSession(). On running the tests, everything is passing 

PS C:\Users\DELL\OneDrive\Documents\kubectl-ai\pkg\sessions> go test -v
=== RUN   TestSessionPersistence
--- PASS: TestSessionPersistence (0.03s)
=== RUN   TestCreateNewSession
--- PASS: TestCreateNewSession (0.02s)
=== RUN   TestDeleteSession
--- PASS: TestDeleteSession (0.00s)
=== RUN   TestListSessions
--- PASS: TestListSessions (0.00s)
=== RUN   TestCorruptedMetadata
--- PASS: TestCorruptedMetadata (0.00s)
=== RUN   TestCorruptedHistory
--- PASS: TestCorruptedHistory (0.01s)
=== RUN   TestConcurrentAccess
--- PASS: TestConcurrentAccess (0.03s)
=== RUN   TestClearMessages
--- PASS: TestClearMessages (0.02s)
=== RUN   TestGetLatestSession
--- PASS: TestGetLatestSession (0.05s)
=== RUN   TestUpdateLastAccessed
--- PASS: TestUpdateLastAccessed (0.04s)
PASS
ok      github.com/GoogleCloudPlatform/kubectl-ai/pkg/sessions  1.462s

nisranjan:
@noahlwest @droot 
The session_test.go did not make it to the kubectl-ai main branch, who need to do something push that to the main branch.
Will closing this PR automatically merge the files into the main branch?

droot:
> @noahlwest @droot The session_test.go did not make it to the kubectl-ai main branch, who need to do something push that to the main branch. Will closing this PR automatically merge the files into the main branch?

Getting this in today. Thank you!

droot:
> > @noahlwest @droot The session_test.go did not make it to the kubectl-ai main branch, who need to do something push that to the main branch. Will closing this PR automatically merge the files into the main branch?
> 
> Getting this in today. Thank you!

presubmits are failing, you may want to run `make check` to make them pass.

PR #506, opened 2025-08-21
feat: meta query tests

ShubyM:
Introduces mocks auto generated by gomock and unit testing of handleMetaQuery


pkg/agent/conversation_test.go
```diff
@@ -0,0 +1,257 @@
...
+					t.Fatalf("unexpected answer: %q", answer)
+				}
+			},
+		},
+	}
+
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			a := tt.expectations(t)
+			ans, handled, err := a.handleMetaQuery(ctx, tt.query)
```
droot, line 242: I remember discussing if we can bring the test driver to closer to how UIs interact with agent.  Are we planning to do that ?

ShubyM: I was planning on making a separate PR for driving the agent in the same way the UI does. I wanted to isolate this PR to introduce mocking and unit testing the handleMetaQuery function

makefile
```diff
@@ -88,6 +88,14 @@ verify-autogen: ## Verify auto-generated files are up to date
 	@echo "λ Verifying auto-generated files..."
 	./dev/ci/presubmits/verify-autogen.sh
 
+generate:
+	go generate ./internal/mocks
+
+verify-mocks:
```
droot, line 94: we should add CI Job like we have for other operations in a follow up. Ask gemini to do it :)

droot (approved):

Issue #449, opened 2025-08-01
[Bug]: `kubectl-ai` downloads binaries without prompting for user permission


bradleydamato:
**Environment (please complete the following):**
- OS: Mac OS Sequoia 15.5
- kubectl-ai version : 0.0.20
- LLM provider: gemini
- LLM model: gemini-2.0-flash

**Describe the bug**
kubectl-ai attempts downloads binaries without prompting for user permission

**To Reproduce**
1. run `kubectl-ai`
2. prompt `>>> check the cluster for any apis that will not be compatible with a k8s 1.29 upgrade`
3. See error

**Expected behavior**
`kubectl-ai` should prompt the user for permission to download binaries before downloading and installing. 

**Additional context**
Allowing the tool to download binaries unprompted is pretty reckless. There is a clear remote code execution risk here. Training data poisoning, AI model capability hiding, or plain user error in prompting could result in full system compromise. It breaks fundamental security boundaries around user intent, and ought to be addressed.

bradleydamato:
<img width="1368" height="515" alt="Image" src="https://github.com/user-attachments/assets/ccc47c4f-d1fd-437a-989b-0049915cba3d" />

justinsb:
Ouch - I thought we asked for approval for commands we didn't know were safe - @droot ?

droot:
> Ouch - I thought we asked for approval for commands we didn't know were safe - [@droot](https://github.com/droot) ?

Yeah, this is bad. We have approval flow for commands that modify the kubernetes resources for protecting k8s cluster but not much approval for running commands on local system. 

So approval for commands and sandbox support, both are needed for production use. 

This is specially needed for using with less powerful models such as `gemini-2.0-flash`. I am wondering what would be the output for `gemini-2.5-pro`.

zvdy:
I wonder if a system prompt for the tool or general would've cached this, similar to my k8s parsing and sysprompt PR to help smaller models.

I would say parsing all these scenarios are unfeasible although curl wget etc are modifying resources

droot:
I researched more on the security aspects of LLM applications and came to conclusion that we can not rely on LLM to determine whether user need to approve a command or not.
Also to prevent exfiltration attacks, asking for approval for composite commands is necessary even if only read-only kubectl commands are involved.

PR: https://github.com/GoogleCloudPlatform/kubectl-ai/pull/510

Thanks @bradleydamato for reporting the issue.

PR #510, opened 2025-08-26
feat: always ask for approval for non-readonly commands

droot:
Two big changes related to command approval flow.

- Do not rely on LLM to determine if a command modifies resource or not.
- If LLM suggests a composite command such as `kubectl get pods | wc -l` or generic bash script command multiple commands, we will always ask for approval even if the kubectl commands are readonly (this is to avoid exfilteration attacks).

## Why
We are seeing requests for using the tool against production clusters and without the above changes, I see a serious risk of prompt injection attacks.

It may appear that the above changes may cause `prompt fatigue` because it will ask the user to approve commands but majority of the use-cases involves read-only kubectl commands and our current static filter will cover those pretty well and not request user to approve those commands.

## Follow ups (coming)
 - Remove the "modifies_resource" parameter in the LLM call completely.
 - Improve the system prompt to steer LLM to avoid using composite calls as much as possible when using kubectl.

fixes https://github.com/GoogleCloudPlatform/kubectl-ai/issues/449

image

### DIFF
--- a/pkg/agent/conversation.go
+++ b/pkg/agent/conversation.go
@@ -931,13 +931,7 @@ func (c *Agent) analyzeToolCalls(ctx context.Context, toolCalls []gollm.Function
 		if err != nil {
 			toolCallAnalysis[i].IsInteractiveError = err
 		}
-		modifiesResourceStr := toolCall.GetTool().CheckModifiesResource(call.Arguments)
-		if modifiesResourceStr == "unknown" {
-			if llmModifies, ok := call.Arguments["modifies_resource"].(string); ok {
-				modifiesResourceStr = llmModifies
-			}
-		}
-		toolCallAnalysis[i].ModifiesResourceStr = modifiesResourceStr
+		toolCallAnalysis[i].ModifiesResourceStr = toolCall.GetTool().CheckModifiesResource(call.Arguments)
 		toolCallAnalysis[i].ParsedToolCall = toolCall
 	}
 	return toolCallAnalysis, nil
--- a/pkg/tools/kubectl_filter.go
+++ b/pkg/tools/kubectl_filter.go
@@ -71,6 +71,7 @@ func kubectlModifiesResource(command string) string {
 
 	hasReadCommand := false
 	foundWrite := false
+	numCmds := 0
 
 	// Single pass through all command calls
 	syntax.Walk(file, func(node syntax.Node) bool {
@@ -87,9 +88,20 @@ func kubectlModifiesResource(command string) string {
 			if result == "no" {
 				hasReadCommand = true
 			}
+			numCmds++
+			if numCmds > 1 {
+				return false // Stop walking if more then one command is found
+			}
 		}
 		return true
 	})
+
+	if numCmds > 1 {
+		// if it's a composite bash command, we should err on the side of caution and return unknown
+		// to prevent exfilteration attacks https://simonwillison.net/2025/Jun/16/the-lethal-trifecta/
+		klog.Infof("KubectlModifiesResource result: unknown for command: %q, multiple commands (%d) found", command, numCmds)
+		return "unknown"
+	}
 
 	// Return results based on what we found
 	if foundWrite {


pkg/agent/conversation.go
```diff
@@ -931,13 +931,7 @@ func (c *Agent) analyzeToolCalls(ctx context.Context, toolCalls []gollm.Function
...
 			toolCallAnalysis[i].IsInteractiveError = err
 		}
-		modifiesResourceStr := toolCall.GetTool().CheckModifiesResource(call.Arguments)
-		if modifiesResourceStr == "unknown" {
-			if llmModifies, ok := call.Arguments["modifies_resource"].(string); ok {
-				modifiesResourceStr = llmModifies
-			}
-		}
-		toolCallAnalysis[i].ModifiesResourceStr = modifiesResourceStr
+		toolCallAnalysis[i].ModifiesResourceStr = toolCall.GetTool().CheckModifiesResource(call.Arguments)
```
janetkuo, line 934: Could it be different modes that users can choose from? e.g. for local / dev, it might be fine to let LLM detect it for me

droot: Some thoughts:

For local/dev where you really don't care about LLM making a mistake, you are ok with the risk, you are probably fine running with `--skip-permissions` flag. right ?

I also realized that in our current flow of delegating to LLM to determine approval requirement is not explicit about it to the end users. I suspect end users expect us to not to make any mistake wrt to approval permission (like https://github.com/GoogleCloudPlatform/kubectl-ai/issues/449) and it surprises them in a big way when it happens. (side comment: our current implementation is still sort anchored around modifying kubernetes resources and we haven't evolved it wrt to custom/mcp tools). 

Thinking more generally, AI agents will have a notion of approval policy with sensible defaults of secure by default and an option to plug-in an external system of approval that they have in their org or delegating it to another system (LLM). [Google's paper](https://arxiv.org/abs/2503.18813) goes it more details around this. [This blog](https://simonwillison.net/2025/Jun/15/ai-agent-security/) also covers agent security in good detail.


So for now I am leaning towards keeping the proposed flow in the PR:

 - Detect readonly operations for kubernetes reliably (0% failure rate) and do not ask for permission to run those operations. 
 - Any other operations (Composite operations even if they involved read-only kubernetes operation to avoid exfiltration attacks [see this for more details on why](https://simonwillison.net/2025/Jun/16/the-lethal-trifecta/)) -- ask the user for approval unless `--skip-permissions` is specified
 - Evolve approval-policy flag to be more pluggable so that we can introduce explicit delegate to LLM or other system over time but making it explicit for the users.

WDYT ?

janetkuo: The tool should not skip permission checks by default. For ease of use, it might be acceptable to skip checks for read-only operations. While this approach might occasionally misclassify some read-only commands (like `kubectl apply --dry-run`) as requiring approval, it is critical that it never fails to catch a write operation.

Even in local/dev environments, I'd prefer not to skip permission checks entirely. If I have high confidence in the LLM, I would be comfortable relying on it for `modifies_resources` detection. That said, I agree that evolving the `approval-policy` flag is a valuable direction.

zvdy: Mentioned this already on the issue itself but we could guardrail the by changing system prompt and making it treat as write operation, my concern is not only kubernetes but more bash and these complex scenarios where we get hallucinations downloading binaries etc

Prompting always too broadly can be not beneficial for UX as simple cat less grep commands can get flagged while debugging an issue, which it's a common scenario

zvdy: For example a system guardrail would be similar to treating any command that writes to disk, downloads or executes a binary, modifies system configuration, or initiates network connections as a mutating operation and always ask for user approval before running.

This is something general that all tools somewhat share so it makes sense to me

droot: @janetkuo 

> While this approach might occasionally misclassify some read-only commands (like kubectl apply --dry-run) as requiring approval,

We handle `--dry-run` flag today and detect that as readonly operation, so we are good there :), but I agree it is critical to that it never fails to detect a write operation.


 > Even in local/dev environments, I'd prefer not to skip permission checks entirely. If I have high confidence in the LLM, I would be comfortable relying on it for modifies_resources detection. That said, I agree that evolving the approval-policy flag is a valuable direction.
 
 Ack.

@zvdy agree on everything you said above.

> Prompting always too broadly can be not beneficial for UX as simple cat less grep commands can get flagged while debugging an issue, which it's a common scenario

This is the part that adds tension and is a trade-off.

> For example a system guardrail would be similar to treating any command that writes to disk, downloads or executes a binary, modifies system configuration, or initiates network connections as a mutating operation and always ask for user approval before running.

Agree. 


I also think there are some creative ways to build approval policy for allowing read/write operations in an incremental ways.

I also think I would like a small LLM that can classify most common shell commands into read/write with 0% failure rate would be amazing :)

janetkuo (approved):
/lgtm

PR #529, opened 2025-09-10
fix: eval hangs

ShubyM:
Looking at the logs with timestamps: https://github.com/GoogleCloudPlatform/kubectl-ai/actions/runs/17626417909/job/50084327013#step:4:154
 
We always appear to hang when trying to see if there are 4 replicas present, so I put a timeout on the request and halved the number we make

### DIFF
--- a/k8s-bench/tasks/scale-down-deployment/setup.sh
+++ b/k8s-bench/tasks/scale-down-deployment/setup.sh
@@ -2,10 +2,10 @@
 # Create namespace and a deployment with initial replicas
 kubectl delete namespace scale-down-test --ignore-not-found
 kubectl create namespace scale-down-test
-kubectl create deployment web-service --image=nginx --replicas=4 -n scale-down-test
+kubectl create deployment web-service --image=nginx --replicas=2 -n scale-down-test
 # Wait for initial deployment to be ready
 for i in {1..30}; do
-    if kubectl get deployment web-service -n scale-down-test -o jsonpath='{.status.availableReplicas}' | grep -q "4"; then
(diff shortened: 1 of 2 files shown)


k8s-bench/tasks/scale-down-deployment/setup.sh
```diff
@@ -2,10 +2,10 @@
 # Create namespace and a deployment with initial replicas
 kubectl delete namespace scale-down-test --ignore-not-found
 kubectl create namespace scale-down-test
-kubectl create deployment web-service --image=nginx --replicas=4 -n scale-down-test
+kubectl create deployment web-service --image=nginx --replicas=2 -n scale-down-test
 # Wait for initial deployment to be ready
 for i in {1..30}; do
-    if kubectl get deployment web-service -n scale-down-test -o jsonpath='{.status.availableReplicas}' | grep -q "4"; then
+    if kubectl --request-timeout=10s get deployment web-service -n scale-down-test -o jsonpath='{.status.availableReplicas}' | grep -q "4"; then
```
noahlwest: If we're making only 2 replicas shouldn't we also grep for 2 here? Also just thinking, adding the timeout of 10s might be enough of a change, since it would cut the time it waits here to 1/3rd (30s default, I believe).

noahlwest (approved):
The change seems fine. I did see some timeouts in the past runs, it looks like maybe 1 out of 6 or so would time out. Is this causing us any problems?

k8s-bench/tasks/scale-down-deployment/setup.sh
ShubyM: Yes fixed to 2! Also I believe the default request timeout is 0 meaning there is no timeout, which I believe is the reason this eval would timeout

Issue #539, opened 2025-09-15
[Feature]: paranoid mode


douglas-gibbons:
**Is your feature request related to a problem? Please describe.**

When working with production systems, it is not always obvious which commands should be confirmed with the user before execution.

For example, an innocent looking read command may result in private customer information being shown.

**Describe the solution you'd like**

Add a flag which can enable users to force _all_ commands to be confirmed before execution

**Describe alternatives you've considered**

Other workarounds, such as use-case specific prompting may work in 99% of cases, but that may not be good enough.

**Additional context**

Too many confirmation requests may just train users to automatically accept everything without first reading what they are agreeing to.

A paranoid mode that could be turned on and off during the conversation may be more beneficial.

droot:
Nice.
I definitely see the use-case for it. This should be very easy to add. Will prioritize it. We can probably introduce a new flag "--approval-policy" with the values "paranoid (always-ask), yolo (never-ask, same as --skip-permission that we have today), auto-approve-read (don't ask for readonly)".

PR #541, opened 2025-09-16
feat: approval policy flag

ShubyM:
Introduces `--approval-policy` flag as described in #539. Paranoid mode also eliminates the choice of saying yes and don't ask me again

What other people do: 

Gemini CLI has very similar options of:
- default: Prompt for approval on each tool call (default behavior)
- auto_edit: Automatically approve edit tools (replace, write_file) while prompting for others
- yolo: Automatically approve all tool calls (equivalent to --yolo)

Claude Code has these 4:

- default: Standard behavior - prompts for permission on first use of each tool
- acceptEdits: Automatically accepts file edit permissions for the session
- plan: Plan Mode - Claude can analyze but not modify files or execute commands
- bypassPermissions: Skips all permission prompts (requires safe environment - see warning below)

### DIFF
--- a/dev/ci/periodics/analyze-evals.sh
+++ b/dev/ci/periodics/analyze-evals.sh
@@ -21,5 +21,12 @@ cd "${REPO_ROOT}/k8s-bench"
 go build -o "${BINDIR}/k8s-bench" .
 
 cd "${REPO_ROOT}"
-"${BINDIR}/k8s-bench" analyze --input-dir "${OUTPUT_DIR}" ${TEST_ARGS:-} -results-filepath ${REPO_ROOT}/.build/k8s-bench.md --output-format markdown
+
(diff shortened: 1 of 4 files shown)


README.md
```diff
@@ -253,7 +253,7 @@ skipVerifySSL: false              # Skip SSL verification for LLM API calls
 
 # Tool and permission settings
 toolConfigPaths: ["~/.config/kubectl-ai/tools.yaml"]  # Custom tools configuration paths
-skipPermissions: false             # Skip confirmation for resource-modifying commands
+approvalPolicy: auto-approve-read # Approval policy: auto-approve-read, paranoid, or yolo
```
droot, line 256: I want to think a bit more on names of these, otherwise the PR is looking good.

Will come back to it tomorrow.

Issue #535, opened 2025-09-12
[Bug]: Kubectl-ai --mcp-client is not able to connect to kubectl-ai --mcp-server running on remote host


iam-ehsan:
**Environment:**
- OS: [e.g. Ubuntu 24.04 lts]

- kubectl-ai version (run `kubectl-ai version`): 
❯ kubectl-ai version
version: 0.0.24
commit: 9889830b6ba8d6082caaffa1fff522655b02adba
date: 2025-09-10T20:33:21Z

- LLM provider:  openai
- LLM model: gpt-4.1

**Describe the bug**
I need to run kubectl-ai as an mcp-server on my remote VM and for that i ran this command *kubectl-ai --mcp-server --mcp-server-mode sse --sse-endpoint-port 9080* and on my local setup i'm running this command *kubectl-ai --mcp-client --quiet "how many pods do i have in all namespaces?"* 

I'm getting various responses depending on my *~/.config/kubectl-ai/mcp.yaml* file 

### 1. If i mention /sse endpoint
```
Servers:
  - Name: kubectl-ai-mcp-server
    URL: http://52.118.249.55:9080/sse
```

```
❯ kubectl-ai --mcp-client --quiet "how many pods do i have in all namespaces?"
E0912 14:57:44.046485   52568 manager.go:118] connecting to MCP server "kubectl-ai-mcp-server": initializing connection: initializing MCP client: transport error: request failed with status 405: Method not allowed

  Running: kubectl get pods --all-namespaces

  I wasn't able to connect to the Kubernetes cluster—I'm seeing a "connection 
  refused" error to localhost:8080. This means kubectl can't access your      
  Kubernetes API server.                                                      
                                                                              
  Possible reasons:                                                           
                                                                              
  • The cluster is not running or is down.                                    
  • Your kubeconfig might not point to a running cluster.                     
  • Networking or firewall issues.                                            
                                                                              
  Please check that your cluster is running and accessible. Let me know once  
  it's resolved, and I'll help you get the pod count in all namespaces!       
```

### 2. If run the curl command it gives me following response and after that i have to use another terminal to execute the commands as the curl command never ends. Have to forcefully stop it, but stopping it also ends the session it created 
 ```
❯ curl http://52.118.249.55:9080/sse
event: endpoint
data: /message?sessionId=b5a44a94-7eae-4902-bc04-9bada71b2c20
```

### 3. Using session generated above in mcp.yaml file, it returns transport error
```
[...]

PR #545, opened 2025-09-24
feat: streamable-http server support

ShubyM:
Currently our MCP server mode does not support the newer [streamable-http](https://modelcontextprotocol.io/specification/2025-06-18/basic/transports#streamable-http) transport. Our client, however, only supports connecting to the newer streamable server, which is why we ran into issues like #535.

This PR adds support for running kubectl-ai as a streamable http server enabling our client to connect to our server. For built in tools this is a bit strange since in this setting both the server and client have bash and kubectl as tools.


droot (approved):
Thank you!

PR #553, opened 2025-10-01
feat: Add ability to create new kind cluster for k8s-bench run

noahlwest:
* adds --create-kind-cluster flag to k8s-bench, which will create a new kind cluster for the invoked run, and delete it at the end

In testing, I noticed rarely that the kind cluster control plane can fail to come up and will close the program.

### DIFF
--- a/k8s-bench/eval.go
+++ b/k8s-bench/eval.go
@@ -26,6 +26,7 @@ import (
 	"regexp"
 	"strings"
 	"sync"
+	"time"
 
 	"github.com/GoogleCloudPlatform/kubectl-ai/k8s-bench/pkg/model"
 	"k8s.io/klog/v2"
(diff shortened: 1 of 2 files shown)


k8s-bench/eval.go
```diff
@@ -26,13 +26,63 @@ import (
...
+	"time"
 
 	"github.com/GoogleCloudPlatform/kubectl-ai/k8s-bench/pkg/model"
 	"k8s.io/klog/v2"
 	"sigs.k8s.io/yaml"
 )
 
 func runEvaluation(ctx context.Context, config EvalConfig) error {
+	if config.CreateKindCluster {
+		clusterName := fmt.Sprintf("k8s-bench-eval-%d", time.Now().Unix())
```
droot: should we fix the name of the cluster so that we can always garbage collect it if it is around? because very often, we will press CTRL-C and that will keep the cluster around if it there is timestamp suffix'd.

k8s-bench/eval.go
```diff
@@ -26,13 +26,63 @@ import (
...
+			return fmt.Errorf("failed to create temp file for kubeconfig: %w", err)
+		}
+		defer os.Remove(kubeconfigFile.Name()) // Clean up the temp file
+
+		if _, err := kubeconfigFile.Write(kubeconfigBytes); err != nil {
+			return fmt.Errorf("failed to write kubeconfig to temp file: %w", err)
+		}
+		kubeconfigFile.Close()
+
+		config.KubeConfig = kubeconfigFile.Name()
```
droot, line 84: We should log the kubeconfig filename (just in case if one wants to manually access to troubleshoot something).

droot (approved):
LGTM, a few minor suggestions.

k8s-bench/eval.go
noahlwest: Sure, logged. Also switched to klog instead of printf.

k8s-bench/eval.go
noahlwest: That's a good point. Initially was thinking we wanted to avoid cluster name clashes so we don't delete anything unintended, but I think we can reasonably assume a fixed name of k8s-bench-eval is intended just for evals and safe to delete; and isn't holding some user's production service.

droot (approved):

PR #556, opened 2025-10-03
feat: Add raw http tracing

noahlwest:
* Adds raw http request and response tracing
\- trace.yaml files can get pretty long due to streaming responses, but the raw response can contain important info so we're logging the whole thing for now
\- currently logs for openai and gemini, though this should work for any other provider that uses an http client if we call withJournaling() there. IIUC, some providers like llamacpp don't use http clients, so we can add a custom logging for any provider that needs it.

* Adds go workspace file
\- `make build-recursive` wouldn't work the journal module wasn't included, even though individual `go build`s would work. This just makes it a little easier and helps with IDE hinting.


pkg/agent/conversation.go
```diff
@@ -220,12 +220,11 @@ func (s *Agent) Init(ctx context.Context) error {
 	systemPrompt, err := s.generatePrompt(ctx, defaultSystemPromptTemplate, PromptData{
 		Tools:             s.Tools,
 		EnableToolUseShim: s.EnableToolUseShim,
```
droot, line 222: SessionIsInteractive got deleted

pkg/agent/conversation.go
```diff
@@ -1044,8 +1029,7 @@ type PromptData struct {
 	Query string
 	Tools tools.Tools
 
-	EnableToolUseShim    bool
-	SessionIsInteractive bool
+	EnableToolUseShim bool
```
droot: SessionIsInteractive ??

pkg/agent/conversation.go
```diff
@@ -290,6 +289,10 @@ func (c *Agent) Close() error {
 func (c *Agent) Run(ctx context.Context, initialQuery string) error {
 	log := klog.FromContext(ctx)
 
+	if c.Recorder != nil {
```
droot, line 294: I think in regular use we do not want http raw trace, but in k8s-bench invocation we do want. 

Can we make this opt-in ? or this already is ?

gollm/http_journal.go
```diff
@@ -0,0 +1,105 @@
...
+		})
+		if writeErr != nil {
+			klog.Errorf("Error writing RoundTripper error to journal: %v", writeErr)
+		}
+		klog.Errorf("RoundTripper error: %v", err)
+		return nil, err
+	}
+
+	// Read the entire response body so we can log it and then pass it along.
+	bodyBytes, err := io.ReadAll(resp.Body)
```
droot, line 66: will it work with the streaming ? or doesn't matter ?

pkg/agent/conversation.go
noahlwest: Looks like I accidentally wrote over #555, this should be fixed now.

pkg/agent/conversation.go
noahlwest: Same thing happened here, accidentally wrote over #555. Fixed now

pkg/agent/conversation.go
noahlwest: for regular use, if tracepath is not supplied, it should just make a dummy logrecorder https://github.com/GoogleCloudPlatform/kubectl-ai/blob/00478f2f7a4eedd2ed877a1bd0ee3982cac48d79/cmd/main.go#L439

Tracepath is provided with k8s-bench always, so no opting out. Should I add a flag for opting in to full http trace and leaving it off by default?

gollm/http_journal.go
noahlwest: It will log each chunk received in streaming, so the output is many lines long, but it does capture everything. I have experimented with some local changes that does a really good job of summarizing up streaming responses, but it loses some data (the *EXACT* content and format that is received). I could try iterating on this, maybe http logger can read the whole response stream, or do something to accumulate it.

PR #557, opened 2025-10-04
feat: fix list-images-for-pods issues

noahlwest:
* Updates list-images-for-pods to no longer use bitnami charts
Due to bitnami removing all their free charts from docker hub, setup script fails every time; https://github.com/bitnami/charts/issues/35164

### DIFF
--- a/gollm/factory.go
+++ b/gollm/factory.go
@@ -193,19 +193,12 @@ func DefaultIsRetryableError(err error) bool {
 // createCustomHTTPClient returns an *http.Client that optionally skips SSL certificate verification.
 // This is shared by all providers that need custom HTTP transport.
 func createCustomHTTPClient(skipVerify bool) *http.Client {
-	if !skipVerify {
-		httpClient := http.DefaultClient
-		httpClient.Timeout = 60 * time.Second
-		return httpClient
-	}
+	transport := http.DefaultTransport.(*http.Transport).Clone()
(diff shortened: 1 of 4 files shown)


gollm/factory.go
```diff
@@ -193,19 +193,12 @@ func DefaultIsRetryableError(err error) bool {
 // createCustomHTTPClient returns an *http.Client that optionally skips SSL certificate verification.
 // This is shared by all providers that need custom HTTP transport.
 func createCustomHTTPClient(skipVerify bool) *http.Client {
-	if !skipVerify {
-		httpClient := http.DefaultClient
-		httpClient.Timeout = 60 * time.Second
-		return httpClient
-	}
+	transport := http.DefaultTransport.(*http.Transport).Clone()
```
droot: Not sure if you want to get this http client related changes in with the `list-*` task changes. Looks like this is related to http tracing.

The change skips setting `Proxy: http.ProxyFromEnvironment` which will disable honoring the http proxy environment variables (a bug).

droot (changes requested):

droot:
The tasks related changes look good. Good catch.

gollm/factory.go
noahlwest: I got my workspace a bit mixed up when trying to break local changes up into different PRs. I just removed that http change, and will also update it in the PR that intends to make this change.

droot (approved):

PR #562, opened 2025-10-06
Feat OpenAI response api

droot:
Command I use to run it:

```sh
export OPENAI_USE_RESPONSES_API=true
./dev/ci/periodics/run-eval-loop.sh 1 openai openai/gpt-oss-20b "http://localhost:8000/v1" "no"
```


droot:
@noahlwest cleaned it up, pl. take a look.

gollm/openai.go
```diff
@@ -48,6 +50,9 @@ func init() {
 	openAIEndpoint = os.Getenv("OPENAI_ENDPOINT")
 	openAIAPIBase = os.Getenv("OPENAI_API_BASE")
 	openAIModel = os.Getenv("OPENAI_MODEL")
+	if os.Getenv("OPENAI_USE_RESPONSES_API") != "" {
```
noahlwest: if I set `export OPENAI_USE_RESPONSES_API=false` this will still use responses api

gollm/openai.go
```diff
@@ -117,6 +122,41 @@ func (c *OpenAIClient) StartChat(systemPrompt, model string) Chat {
...
+				Model:           selectedModel,
+				Temperature:     openai.Float(0.2),
+				MaxOutputTokens: openai.Int(2048),
+				Include:         []responses.ResponseIncludable{
+					// responses.ResponseIncludableReasoningEncryptedContent,
+				},
+				Reasoning: responses.ReasoningParam{
+					Effort: responses.ReasoningEffortLow,
+				},
+				Store: openai.Bool(false),
```
noahlwest, line 154: Not sure what the ResponseIncludableReasoningEncryptedContent comment is for.

Also, how did you decide on these values? Do we want to make them configurable (maybe in a later PR?) I see, for example, gemini 2.5 pro has 65,535 max output tokens https://cloud.google.com/vertex-ai/generative-ai/docs/models/gemini/2-5-pro GPT-5 has 128,000 max output tokens https://platform.openai.com/docs/models/gpt-5. Maybe we should up it a bit?

gollm/openai_response.go
```diff
@@ -0,0 +1,341 @@
...
+	}
+	klog.V(1).Infof("Set %d function definitions for OpenAI chat session", len(cs.functionDefinitions))
+	return nil
+}
+
+// Send sends the user message(s), appends to history, and gets the LLM response.
+func (cs *openAIResponseChatSession) Send(ctx context.Context, contents ...any) (ChatResponse, error) {
+	klog.V(1).InfoS("openAIChatSession.Send called", "model", cs.model, "history_len", len(cs.history))
+
+	return &openAIChatResponse{}, errors.ErrUnsupported
```
noahlwest, line 77: We should add a comment here about why we don't support this.

gollm/openai_response.go
```diff
@@ -0,0 +1,341 @@
...
+	}
+	var candidates []Candidate
+	for _, output := range r.resp.Output {
+		switch output.AsAny().(type) {
+		case responses.ResponseFunctionToolCall, responses.ResponseOutputMessage:
+			candidates = append(candidates, &openAIResponseCandidate{
+				candidate: &output,
+			})
+		default:
+			// skip reasoning messages because agentic loop doesn't know
```
noahlwest, line 175: should we log this?

noahlwest (changes requested):

gollm/openai.go
droot: I was playing around with these values, not needed for our use case, so removed it (reasoning-encrypted-content).

We need to figure out how to support LLM specific configuration. Some configurations are common and some are not. Deferring it for now.

gollm/openai_response.go
droot: Added.

gollm/openai_response.go
droot: This is logged before when we receive the response, so not needed here.

gollm/openai.go
droot: Fixed. tx.

droot:
@noahlwest thanks for the review. Addressed the feedback and verified it as well. Ready for final review.

gollm/openai.go
noahlwest: sgtm. I've also tested locally with these values and the outputs seem good and relatively consistent.

PR #563, opened 2025-10-07
feat: add eval multi-run dev script

noahlwest:
Script runs k8s-bench, analyzes, adds the timer to markdown output, and writes all results to a numbered output folder.

Example usage: ./run-eval-loop.sh --iterations 5 --provider openai --model Qwen/Qwen3-Next-80B-A3B-Instruct --api-base http://localhost:8000/v1 --concurrency 5 --task-pattern "create"

### DIFF
--- a/dev/ci/periodics/run-eval-loop.sh
+++ b/dev/ci/periodics/run-eval-loop.sh
@@ -1,15 +1,46 @@
 #!/bin/bash
 
+set -eou pipefail
+
 # Positional args
-# 1. ITERATIONS: Number of times to run the loop (default: 3)
-# 2. PROVIDER:   The LLM provider to use (default: openai)
-# 3. MODEL:      The specific model to test (default: "Qwen/Qwen3-Next-80B-A3B-Instruct")
-# 4. API_BASE:   The API base URL (default: "http://localhost:8000/v1")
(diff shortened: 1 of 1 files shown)


dev/ci/periodics/run-eval-loop.sh
```diff
@@ -0,0 +1,163 @@
...
+echo "Model:     $MODEL"
+echo "API Base:  $API_BASE"
+echo "Concurrency: $CONCURRENCY"
+echo "Task Pattern: ${TASK_PATTERN:-"All Tasks"}"
+
+# Loop from 1 to the specified number of iterations
+for i in $(seq 1 $ITERATIONS)
+do
+  # Create the unique directory name for this run
+  OUTPUT_DIR="${REPO_ROOT}/.build/k8s-bench-${MODEL}-${i}"
```
droot, line 95: If the model is `openai/gpt-oss-20' (containing file delimiter), the output_dir will be `./build/k8s-bench-openai/gpt-oss-20b-1` which is probably not what we want. May be for deriving directory path, we can replace the delimiter with something else say `-` or something else ?

noahlwest: Yeah the generated structure is a bit less predictable because of that, we can adjust it in a new pr since this one is submitted.

PR #566, opened 2025-10-07
evals: Add per-task timeout to k8s-bench

noahlwest:
Adds default 5 minute timeout to any task. Per-task timeout can also be set in the task.yaml

### DIFF
--- a/k8s-bench/eval.go
+++ b/k8s-bench/eval.go
@@ -26,6 +26,7 @@ import (
 	"regexp"
 	"strings"
 	"sync"
+	"time"
 
 	"github.com/GoogleCloudPlatform/kubectl-ai/k8s-bench/pkg/model"
 	"k8s.io/klog/v2"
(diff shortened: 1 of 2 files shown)


k8s-bench/eval.go
```diff
@@ -297,20 +312,25 @@ func evaluateTask(ctx context.Context, config EvalConfig, taskID string, task Ta
...
-	if err := x.runSetup(ctx); err != nil {
+	if err := x.runSetup(taskCtx); err != nil {
 		// Unexpected error
 		result.Error = err.Error()
 		return result
 	}
 
 	// Run the agent
-	agentOutput, err := x.runAgent(ctx)
+	agentOutput, err := x.runAgent(taskCtx)
```
droot, line 327: I am wondering reusing the same ctx is a good idea ? will this stack up the timeouts ? or will the overall setup/run will have a deadline of 5 mins or individually 5 min. deadline ?

noahlwest: This should have a deadline of 5 minutes total per eval (create-canary gets 5m, fix-image-pull gets 5m, etc.) and the 5 minute limit is for the setup, run, and cleanup. We could up it to 10 to be safer since it includes the setup, run, and cleanup.

droot (approved):

PR #569, opened 2025-10-14
feat: pluggable agent support in k8s-bench

ShubyM:


droot:
@ShubyM sorry the delay, will get to it tomorrow.

@noahlwest pl. take a look at it tomorrow as well.

dev/ci/periodics/run-eval-loop.sh
noahlwest, line 1: I wasn't able to run this shell script with your change as is. I ended up removing the kubeconfig flag from `run-evals.sh` and was able to run it, but I also ended up changing the `make run-evals` line in `run-eval-loop.sh` to instead just directly run k8s-bench, and added the flags it needed in TEST_ARGS.

```
run_time_line=$( \
    OPENAI_API_KEY="not needed" \
    OPENAI_API_BASE="$API_BASE" \
    TEST_ARGS="$TEST_ARGS" \
    $K8S_BENCH_BIN run $TEST_ARGS --agent-bin=$(which gemini) | tee /dev/tty | grep '^Total evaluation time:' \
  )
```
and had it running against gemini. Though, I don't think this is what you had in mind, since agent-args would have to be removed to work with gemini. I'm interested in how you had things set up using this script, or if it was just with k8s-bench directly

k8s-bench/README.md
```diff
@@ -48,13 +52,13 @@ The `run` subcommand executes the benchmark evaluations.
 | `--agent-bin` | Path to kubectl-ai binary | - | Yes |
 | `--output-dir` | Directory to write results to | - | Yes |
 | `--tasks-dir` | Directory containing evaluation tasks | ./tasks | No |
-| `--kubeconfig` | Path to kubeconfig file | ~/.kube/config | No |
 | `--task-pattern` | Pattern to filter tasks (e.g. 'pod' or 'redis') | - | No |
-| `--llm-provider` | Specific LLM provider to evaluate (e.g. 'gemini' or 'ollama') | gemini | No |
-| `--models` | Comma-separated list of models to evaluate | gemini-2.5-pro-preview-03-25 | No |
-| `--enable-tool-use-shim` | Enable tool use shim | false | No |
-| `--quiet` | Quiet mode (non-interactive mode) | true | No |
+| `--agent-args` | Additional argument passed directly to the agent (repeat flag to provide multiple) | - | No |
```
noahlwest, line 56: We lose some info here, we could include a pointer where people can see the different options they'd add here (llm-provider, models, etc.)

k8s-bench/eval.go
```diff
@@ -149,8 +153,18 @@ func runEvaluation(ctx context.Context, config EvalConfig) error {
 						log = logFile
 					}
 
+					start := time.Now()
+					fmt.Printf("\033[36mWorker %d: Started %s for %s\033[0m\n", workerID, llmConfig.ID, job.taskID)
```
noahlwest, line 157: I was a little confused by this at first, it changes the color in terminal. Maybe we should add a comment pointing that out?

I like that it's easier to read, though I wonder how it shows up in the other interfaces (reminds me of the #384). Probably it's no big deal, I don't imagine many people are running k8s-bench in tui or web ui. 

image

k8s-bench/eval.go
```diff
@@ -149,8 +153,18 @@ func runEvaluation(ctx context.Context, config EvalConfig) error {
...
+
 					result := evaluateTask(ctx, config, job.taskID, job.task, llmConfig, log)
 
+					fmt.Printf("\033[32mWorker %d: Completed %s for %s in %s\033[0m\n",
+						workerID,
+						llmConfig.ID,
+						job.taskID,
+						time.Since(start).Round(time.Second),
```
noahlwest, line 165: This is pretty nice to have!

k8s-bench/main.go
```diff
@@ -206,10 +210,104 @@ func (f *Strings) String() string {
...
 }
 
+type agentDetector func(agentBin string, agentArgs []string, cfg *model.LLMConfig) bool
+
+var agentDetectors = []agentDetector{
+	detectGeminiAgent,
+}
+
+func detectAgentLLMConfig(agentBin string, agentArgs []string) model.LLMConfig {
+	llmConfig := model.LLMConfig{
```
noahlwest, line 230: This ends up losing us some data, ex:
```
error: ""
llmConfig:
  enableToolUseShim: false
  id: agent
  model: default
  provider: agent
  quiet: false
name: create-pod-mount-configmaps
result: success
```

I got this in the results.yaml after a run, not really helpful for the id, model and provider.

k8s-bench/eval.go
```diff
@@ -515,25 +539,25 @@ func (x *TaskExecution) runCleanup(ctx context.Context) error {
 }
 
 func (x *TaskExecution) runAgent(ctx context.Context) (string, error) {
-	tracePath := filepath.Join(x.taskOutputDir, "trace.yaml")
```
noahlwest, line 537: I think we should add this back, I'm not sure that we can have individual traces for each task without it.

noahlwest (changes requested):
It would be helpful if your pr had some instructions on any necessary setup, examples commands of running (with kubectl-ai and gemini), and any explanation you feel would be helpful to anyone who looks at this pr.

dev/ci/periodics/run-eval-loop.sh
ShubyM: Oh yea I was running with k8s directly, I can adapt the script

PR #570, opened 2025-10-14
enhance: canary deployment task

ShubyM:
Saw that gemini-cli was failing this task because it would create the new deployment inside the default namespace. Introducing the new namespace in the prompt fixes the issue. Also reduced replicas to two replicas to lessen pressure on the cluster


noahlwest (approved):
Overall, this is a good change that will increase the consistency of this eval. However: I'm wondering how much information we should give in the prompts. Maybe it's worth having some evals that are purposely very ambiguous or very casually prompted to see how well models can decipher what we want, and some other evals that precisely describe the task to see how well models can follow detailed instructions.

droot (approved):

droot:
> Overall, this is a good change that will increase the consistency of this eval. However: I'm wondering how much information we should give in the prompts. Maybe it's worth having some evals that are purposely very ambiguous or very casually prompted to see how well models can decipher what we want, and some other evals that precisely describe the task to see how well models can follow detailed instructions.

+1

If we look at the eval, agent is being tested on two things.

- for an under specified problem, is the agent able to discover the additional information about the issue before doing the actual task
- Doing the task. The task here is about agent knowing how to do canary deployment.

With the changes, focus of the eval now shifts to performing the actual task which is good but I am still thinking how to eval the agent's behavior for an under specified problem. Agent that tries to go ahead and creates deployment in default namespace without examining the cluster state is actually bigger failure in my opinion :( and probably should be captures as negative tests. If there was human in the loop, agent prompting human for more information is a positive sign (better behavior will be agent doing a lightweight discovery and then confirming from the user where user can simply say yes/no and correct it.)

But overall, I like introducing category of evals that tests how agent helps ambiguous information and keeping them separate from the functional eval.

PR #572, opened 2025-10-15
eval: Generally up timeouts for verify/setup scripts

noahlwest:
We're seeing some timeout issues in eval runs, this change mostly ups timeout lengths as an experiment to see if it helps. A few evals also get updates to setup/verify scripts.
* fix-oomkilled was passing for no reason. It still can, if the setup doesn't create oomkilled events, but this at least checks that memory was changed
* fix-pending-pod scripts give us a little bit more visibility into what's happening
* same with rolling-update-deployment

### DIFF
--- a/k8s-bench/tasks/create-canary-deployment/verify.sh
+++ b/k8s-bench/tasks/create-canary-deployment/verify.sh
@@ -12,7 +12,7 @@ CANARY_IMAGE="nginx:1.29"
 CANARY_REPLICAS=2
 STABLE_REPLICAS=2
 EXPECTED_TOTAL_ENDPOINTS=$((CANARY_REPLICAS + STABLE_REPLICAS))
-TIMEOUT="30s"
+TIMEOUT="120s"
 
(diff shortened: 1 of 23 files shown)


k8s-bench/tasks/create-canary-deployment/verify.sh
```diff
@@ -12,7 +12,7 @@ CANARY_IMAGE="nginx:1.29"
 CANARY_REPLICAS=1
 STABLE_REPLICAS=10
 EXPECTED_TOTAL_ENDPOINTS=$((CANARY_REPLICAS + STABLE_REPLICAS))
-TIMEOUT="30s"
+TIMEOUT="120s"
```
droot, line 15: Was chatting offline with @ShubyM and we decided to decrease the replicas from 10 to 2. https://github.com/GoogleCloudPlatform/kubectl-ai/pull/570

I suspect high resource requirement might be reason for things to take more time especially in high concurrency setting.

PR #574, opened 2025-10-15
k8sbench: added cluster-creation-policy support

droot:
k8sbench offers more control around the target cluster creation for running the eval.


droot:
@noahlwest can you pl. take a look at this one.

PR #580, opened 2025-10-16
k8s-bench: Added support for task filtering using tags

droot:
Used gemini-cli to see how we can add tag based task filtering.

/cc @noahlwest

### DIFF
--- a/dev/ci/periodics/run-eval-loop.sh
+++ b/dev/ci/periodics/run-eval-loop.sh
@@ -16,6 +16,8 @@ API_BASE="http://localhost:8000/v1"
 CONCURRENCY=1
 # The regex pattern for tasks to run
 TASK_PATTERN=""
+# The tags for tasks to run
+TASK_TAGS=""
 # kind cluster creation policy (default: "CreateIfNotExists")
 CLUSTER_CREATION_POLICY="CreateIfNotExists"
 
@@ -44,6 +46,10 @@ while [[ $# -gt 0 ]]; do
       ;;
     -t|--task-pattern)
       TASK_PATTERN="$2"
+      shift 2
+      ;;
+    -T|--task-tags)
+      TASK_TAGS+="--task-tags=$2 "
       shift 2
       ;;
     -k|--cluster-creation-policy)
@@ -93,6 +99,7 @@ echo "Model:     $MODEL"
 echo "API Base:  $API_BASE"
 echo "Concurrency: $CONCURRENCY"
 echo "Task Pattern: ${TASK_PATTERN:-"All Tasks"}"
+echo "Task Tags: ${TASK_TAGS:-"No Tags"}"
 
 # Loop from 1 to the specified number of iterations
 for i in $(seq 1 $ITERATIONS)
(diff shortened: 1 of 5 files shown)


dev/ci/periodics/run-eval-loop.sh
```diff
@@ -46,6 +48,10 @@ while [[ $# -gt 0 ]]; do
       TASK_PATTERN="$2"
       shift 2
       ;;
+    -T|--task-tags)
+      TASK_TAGS+="--task-tags=$2 "
```
noahlwest, line 52: I tested this out:
This results in "--task-tags=cluster=gke" being passed if I use `--task-tags cluster=gke`. This gets incorrectly parsed and tasks with the tag don't get ran.

TASK_TAGS+="$2 " results in correct cluster=gke, which gets correctly parsed and runs the tasks we want, but doesn't work if I add some other filter e.g., `--task-tags cluster=gke --task-tags api=hpa`, will only filter by cluster=gke.

k8s-bench/eval.go
```diff
@@ -242,12 +242,59 @@ func loadTasks(config EvalConfig) (map[string]Task, error) {
...
 		tasks[taskID] = task
 	}
 
 	return tasks, nil
 }
 
+// matchTags checks if a task matches the provided tag filters.
+// If a task has no tags, it is always included.
+// If filters are provided, a task with tags must match them.
+func matchTags(task Task, filters []string) bool {
```
noahlwest, line 258: I pulled this to test and this behaves differently that I expect. I'm including -t (--task-pattern) to select some tasks with and some without tags, some test scenarios:

If we wanted to run all tasks, we don't include any --task-tags arg. For example:
`./dev/ci/periodics/run-eval-loop.sh -i 1 -p gemini -m gemini-2.5-pro -a http://localhost:8000/v1 -c 5 -k AlwaysCreate -t "resize-pvc|horizontal-pod-autoscaler|debug-app-logs"`

If we wanted to run all tasks, except those tagged with gke, we include ANY filter:
`... -t "resize-pvc|horizontal-pod-autoscaler|debug-app-logs" --task-tags fake=tag`
This only runs debug-app-logs task

If I wanted to run (only) tasks tagged with gke, because we already ran everything else on KinD, we can't?
`... -t "resize-pvc|horizontal-pod-autoscaler|debug-app-logs" --task-tags cluster=gke`
That will run resize, hpa, and debug tasks.

Also if we have more tags, and only want to exclude cluster=gke tasks, we can't do that without including every single other tag?
`... -t "resize-pvc|horizontal-pod-autoscaler|debug-app-logs" --task-tags api=hpa,label2=value2` etc...

Maybe we'd want something similar to k8s testing, where they have a skip tag and we provide regex pattern, but for tags? --skip-tags "gke" or --skip-tags"NotWithKind"

Right now, we can do something like `-t "^(?!(horizontal-pod-autoscaler|resize-pvc)$).*$"` to avoid the specific tests, which doesn't require any extra code, just making the right regex pattern.

droot:
Thanks @noahlwest for the review. I think it makes more sense to close this PR and re-do it properly if we need it in the future.

PR #593, opened 2025-10-21
eval: remove hints from crashloop and probes

ShubyM:
Changed fix-crashloop and fix-probes to remove hints


droot:
Will defer it to @noahlwest

droot:
Took another look and the change is very reasonable, so getting this in.

PR #594, opened 2025-10-22
feat: Return unexpected kubectl-ai errors in RunOnce mode, display them in k8s-bench

noahlwest:
* kubectl-ai returns unexpected errors encountered in kubectl-ai RunOnce mode
* k8s-bench run now counts unexpected errors as errors instead of failures
* k8s-bench analyze now accounts for errors encountered

Tested this with local served gpt-oss-20b, errors like 404, 400, tool parsing errors, etc. are saved and recorded as errors now in k8s-bench. Easiest way to reliably recreate errors is to use start a vllm instance and point to a model that doesn't exist, or use a model with known issues (like gpt-oss-20b on completions endpoint.)

* When unexpected error is encountered and returned to k8s-bench, we print usage/help text for kubectl-ai. This adds a little bit of noise to the logs but we can fix that in a later pr

### DIFF
--- a/k8s-bench/eval.go
+++ b/k8s-bench/eval.go
@@ -328,7 +328,25 @@ func evaluateTask(ctx context.Context, config EvalConfig, taskID string, task Ta
 			return result
 		}
 		// Unexpected error
-		result.Error = err.Error()
+		result.Result = "error"
+		const maxErrLogLines = 3
+		logString := logBuffer.String()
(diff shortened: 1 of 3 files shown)


droot (approved):

droot:
Thank you so much for quick turn around on this one. This addresses a critical issues.

PR #611, opened 2025-11-19
gollm: tolerate empty streaming parts in Gemini responses

justinsb:
This seems to happen (more?) with gemini-3-pro-preview models.


noahlwest:
Pulled this locally and gave it some manual testing, as well as ran it against k8s-ai-bench. Looks to be running well with gemini 3 and 2.5, haven't seen any issues yet.

PR #613, opened 2025-11-20
feat: use-sandbox and use-image flag

ShubyM:
This PR is  adds support for sandboxing against a GKE cluster and macOS seatbelt.The code for sandboxing against the cluster matches #433. I also introduced an abstraction called Executor that enables us to separate logic between the responsibility of tools vs the environment they actually run in.

### DIFF
--- a/cmd/main.go
+++ b/cmd/main.go
@@ -128,7 +128,7 @@ type Options struct {
 	ShowToolOutput bool `json:"showToolOutput,omitempty"`
 
 	// Sandbox enables execution of tools in a sandbox environment.
-	// Supported values: "cluster", "mac".
+	// Supported values: "k8s", "seatbelt".
 	// If empty, tools are executed locally.
 	Sandbox string `json:"sandbox,omitempty"`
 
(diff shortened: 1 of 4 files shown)


pkg/tools/custom_tool.go
```diff
@@ -137,12 +137,18 @@ func (t *CustomTool) Run(ctx context.Context, args map[string]any) (any, error)
 	}
 
 	workDir := ctx.Value(WorkDirKey).(string)
+	env := os.Environ()
+
+	// Get executor from context or default to local
+	var executor sandbox.Executor
+	if v := ctx.Value(ExecutorKey); v != nil {
```
droot, line 144: I was wondering if we should add `executor` field to the `customTool` struct like we did for bashTool and kubectlTool and avoid getting the executor from the ctx ?

droot (approved):
LGTM. I have a minor comment.

pkg/tools/custom_tool.go
droot: I thought more about it. `CustomTool` aren't that popular because `bashTool` works well to invoke any CLI tool. Let's ignore this for now.

PR #615, opened 2025-12-01
Add k8s-ai-bench site to present benchmark results

noahlwest:
generate_benchmark.py : generates the pages to display k8s-ai-bench data
combined_results.jsonl : the benchmark data that is used while generating the site
about.html : a quick about page to describe k8s-ai-bench
index.html : main page, shows the model leaderboard and links out to drilled-down views
model.html : the drilled down view from the main page, shows the info from a single mode's runs. NOTE: the "view full log" button is currently disabled, so this will only show the last few lines of a failure log.
tasks.html : tasks overview page, ranks tasks by difficulty, links out to their source code, and offers a drill-down to a single-task view that shows how all the models performed on that task
task_detail.html : the drilled down view from the tasks page, shows how all the models did for the particular task and links out to the model page that shows all the task runs for that model.

Here is a quick click through of the pages:
https://github.com/user-attachments/assets/8f92528a-76ed-483e-83dd-83a3f08fa6dd

### DIFF
--- a/k8s-ai-bench/README.md
+++ b/k8s-ai-bench/README.md
@@ -161,6 +161,23 @@ The `analyze-evals.sh` script analyzes the results from the previous run.
 
 The results will be saved to `.build/k8s-ai-bench.md` and `.build/k8s-ai-bench.json`.
 
+### Visualization
+
+The `site` directory contains a static website for visualizing the benchmark results.
+
+To view the visualization locally:
+
+1.  Ensure you have a `combined_results.jsonl` file in the `site` directory. You can generate this by running the `analyze` command with the `jsonl` output format:
+    ```sh
+    ./k8s-ai-bench analyze --input-dir .build/k8s-ai-bench --output-format jsonl --results-filepath site/combined_results.jsonl
+    ```
+2.  Serve the site locally (browsers block `file://` access to local files for security reasons):
+    ```sh
+    cd site
+    python3 -m http.server
+    ```
+3.  Open [http://localhost:8000](http://localhost:8000) in your browser.
+
 ### Contributions
 
 We're open to contributions in k8s-ai-bench, check out the [contributions guide.](contributing.md)
--- a/k8s-ai-bench/site/style.css
+++ b/k8s-ai-bench/site/style.css
@@ -1,50 +1,350 @@
-body { font-family: -apple-system, BlinkMacSystemFont, "Segoe UI", Roboto, Helvetica, Arial, sans-serif; margin: 0; padding: 0; background-color: #f6f8fa; color: #24292e; display: flex; flex-direction: column; min-height: 100vh; }
-.navbar { background-color: #24292e; padding: 1rem 2rem; display: flex; align-items: center; justify-content: space-between; box-shadow: 0 2px 4px rgba(0,0,0,0.1); }
-.nav-brand { color: white; font-weight: bold; font-size: 1.2rem; text-decoration: none; margin-right: 2rem; }
-.nav-links { display: flex; gap: 1rem; }
-.nav-btn { color: rgba(255,255,255,0.7); text-decoration: none; font-weight: 600; padding: 0.5rem 1rem; border-radius: 6px; transition: all 0.2s; font-size: 0.9rem; }
-.nav-btn:hover { color: white; background-color: rgba(255,255,255,0.1); }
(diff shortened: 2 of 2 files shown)


k8s-ai-bench/site/generate_benchmark.py
```diff
@@ -0,0 +1,689 @@
+import json
```
droot: Curious, any specific reason for using python over Go for this ?

Given that we have Go code for analyzing the eval runs, it will be nice to keep all the tooling in one language. Thoughts ?

k8s-ai-bench/site/generate_benchmark.py
```diff
@@ -0,0 +1,689 @@
...
+        
+        # Sort breakdown by P1 desc
+        t_breakdown.sort(key=lambda x: x['p1'], reverse=True)
+        task_details_map[t_name] = t_breakdown
+
+    # Final Sorting
+    leaderboard.sort(key=lambda x: x['p5'], reverse=True)
+    task_stats_summary.sort(key=lambda x: x['p1']) # Hardest first (lowest score)
+
+    return {
```
droot: I was wondering if we should avoid this aggregation and instead have just flat task run in `.jsonl`. Aggregation can be done by the JS functions in the javascript part of the frontend.

Main reason being making it easy to append new model runs. New runs will be simply adding rows for new task runs. Thoughts ?

k8s-ai-bench/site/generate_benchmark.py
```diff
@@ -0,0 +1,689 @@
...
+        "leaderboard": leaderboard,
+        "tasks": task_stats_summary,
+        "details": model_details_map,
+        "task_details": task_details_map
+    }
+
+# --- 2. HTML GENERATION HELPERS ---
+
+def get_common_head(title, data_json_str=None):
```
droot: Few thoughts:

I was wondering if we can structure frontend and data (jsonl) such that we don't need to generate these.

The static files can assume that the `jsonl` file can be fetched from`/combined_...json` relative to wherever static files are hosted on.

The decoupling will make it easy to iterate on using AI as well. 

If we want to support self-contained static files, we use Go's html template package to switch between loading the `jsonl` or embedding inline.

k8s-ai-bench/site/generate_benchmark.py
noahlwest: No reason to use python over go, was just out of habit. I've switched it over to a go program instead.

k8s-ai-bench/site/generate_benchmark.py
noahlwest: Agreed that this is better. I've made the change so the data processing is on the js side, and also verified that to add new runs we only need to add to the jsonl file and refresh.

k8s-ai-bench/site/generate_benchmark.py
noahlwest: This is updated now so the pages fetch the jsonl file

k8s-ai-bench/site/generate_benchmark.go
```diff
@@ -0,0 +1,568 @@
+package main
```
droot: Now that the data is de-coupled from the HTML files, do we really need the site generator ? We can simply have these HTML files as static files. Another big advantage of having these static files is that we can iterate on the HTML files using AI IDEs rapidly.

Also, for generating HTML files, Go has html and text template package, that makes it easy to manage these HTML files. the HTML template also supports some nice features such as embedding the template etc.

k8s-ai-bench/site/generate_benchmark.py
droot: After the decoupling of data from the html files, I do not see a good reason for site generator. I have ideas on how we can do that. Scheduled GVC to go over that.

k8s-ai-bench/site/generate_benchmark.py
droot: Thanks.

k8s-ai-bench/site/generate_benchmark.go
noahlwest: I've removed the generator and instead structured the files more like a normal site, since all the generator was doing was holding common styles and js.

droot (approved):

PR #616, opened 2025-12-03
Unify session backends 

ShubyM:
Introduced the concept of a "store" to unify logic among different storage backends. Also introduced JSONL format into stored sessions on filesystem example below:

```json
{"ID":"86501a38-3854-4cb6-b2c2-f0a9cd970d6f","Source":"agent","Type":"text","Payload":"Hey there, what can I help you with today?","Timestamp":"2025-12-03T20:02:41.36281053Z"}
{"ID":"ec618a3f-e4da-4149-ac6a-3dcca6c1c5a0","Source":"agent","Type":"user-input-request","Payload":"\u003e\u003e\u003e","Timestamp":"2025-12-03T20:02:41.363389637Z"}
{"ID":"adc665de-9c8d-4622-8354-b95ddeabedbd","Source":"user","Type":"text","Payload":"Can you tell me about my cluster?","Timestamp":"2025-12-03T20:03:07.786590033Z"}
{"ID":"ea919e5f-dc23-4399-84a9-39d8548d0c77","Source":"model","Type":"text","Payload":"Of course! I can give you a summary of your Kubernetes cluster. Let's start by getting some basic information. 🧐\n\nFirst, I'll check the general cluster information, like the Kubernetes version and the control plane endpoint.\n\n`kubectl cluster-info` is a good starting point. After that, I'll look into the nodes that make up your cluster.\n\nLet's begin.","Timestamp":"2025-12-03T20:03:12.270462562Z"}
{"ID":"2d405551-dec5-484b-94db-6c901c9a2d31","Source":"agent","Type":"user-input-request","Payload":"\u003e\u003e\u003e","Timestamp":"2025-12-03T20:03:12.270750586Z"}
```

### DIFF
--- a/pkg/agent/conversation.go
+++ b/pkg/agent/conversation.go
@@ -426,11 +426,11 @@ func (c *Agent) Run(ctx context.Context, initialQuery string) error {
 		} else {
 			if len(c.Session.Messages) > 0 {
 				// Resuming existing session
-				greetingMessage := "Welcome back. What can I help you with today?\n (Don't want to continue your last session? Use --new-session)"
+				greetingMessage := fmt.Sprintf("Welcome back. What can I help you with today?\n (Don't want to continue your last session? Use --new-session)\n\n%s", c.Session.String())
 				c.addMessage(api.MessageSourceAgent, api.MessageTypeText, greetingMessage)
 			} else {
 				// Starting new session
-				greetingMessage := "Hey there, what can I help you with today?"
+				greetingMessage := fmt.Sprintf("Hey there, what can I help you with today?\n\n%s", c.Session.String())
 				c.addMessage(api.MessageSourceAgent, api.MessageTypeText, greetingMessage)
 			}
 		}
@@ -811,15 +811,7 @@ func (c *Agent) handleMetaQuery(ctx context.Context, query string) (answer strin
 	case "tools":
 		return "Available tools:\n\n  - " + strings.Join(c.Tools.Names(), "\n  - ") + "\n\n", true, nil
 	case "session":
-		sessionInfo := fmt.Sprintf(
-			"Current session:\n\nID: %s\nCreated: %s\nLast Accessed: %s\nModel: %s\nProvider: %s\n\n",
-			c.Session.ID,
-			c.Session.CreatedAt.Format("2006-01-02 15:04:05"),
-			c.Session.LastModified.Format("2006-01-02 15:04:05"),
-			c.Session.ModelID,
-			c.Session.ProviderID,
-		)
-		return sessionInfo, true, nil
+		return c.Session.String(), true, nil
 
 	case "save-session":
 		savedSessionID, err := c.SaveSession()
--- a/pkg/api/models.go
+++ b/pkg/api/models.go
@@ -15,6 +15,7 @@
 package api
 
 import (
+	"fmt"
 	"time"
 )
 
@@ -128,4 +129,9 @@ func (s *Session) AllMessages() []*Message {
 		return nil
 	}
 	return s.ChatMessageStore.ChatMessages()
+}
+
+func (s *Session) String() string {
+	return fmt.Sprintf("Session ID: %s\nProvider: %s\nModel: %s\nCreated At: %s\nLast Modified: %s\nAgent State: %s",
+		s.ID, s.ProviderID, s.ModelID, s.CreatedAt.Format(time.RFC3339), s.LastModified.Format(time.RFC3339), s.AgentState)
 }


pkg/agent/conversation.go
```diff
@@ -207,18 +213,38 @@ func (s *Agent) Init(ctx context.Context) error {
 		return fmt.Errorf("RunOnce mode requires an initial query to be provided")
 	}
 
-	s.session = &api.Session{
-		Messages:         s.ChatMessageStore.ChatMessages(),
-		AgentState:       api.AgentStateIdle,
-		ChatMessageStore: s.ChatMessageStore,
+	if s.SessionBackend == "" {
```
droot, line 209: With the re-factoring, we should ensure that Agent is always created with some session (default to memory) and that should be done at the time of Agent creation. 

That should eliminate the need for the logic of creating a session backed etc. in the `addMessage` method.

pkg/agent/conversation.go
```diff
@@ -149,6 +149,12 @@ func (s *Agent) Session() *api.Session {
 	return &sessionCopy
 }
 
+func (s *Agent) SetSession(session *api.Session) {
```
droot: I wonder why is this needed.

pkg/agent/conversation.go
```diff
@@ -785,14 +811,18 @@ func (c *Agent) handleMetaQuery(ctx context.Context, query string) (answer strin
 	case "tools":
 		return "Available tools:\n\n  - " + strings.Join(c.Tools.Names(), "\n  - ") + "\n\n", true, nil
 	case "session":
-		if s, ok := c.ChatMessageStore.(*sessions.Session); ok {
-			out, err := s.String()
-			if err != nil {
-				return "", false, fmt.Errorf("failed to get session string: %w", err)
-			}
-			return out, true, nil
+		if c.session == nil {
```
droot: we should never be in this state given that we will always have a session (defaulted to memory).

pkg/agent/conversation.go
```diff
@@ -785,14 +811,18 @@ func (c *Agent) handleMetaQuery(ctx context.Context, query string) (answer strin
...
+			return "session not found (session persistence not enabled)", true, nil
 		}
-		return "Session not found (session persistence not enabled)", true, nil
+		sessionInfo := fmt.Sprintf(
```
droot: I liked the logic of using `s.String()` to print details of session.

pkg/sessions/filesystem.go
```diff
@@ -0,0 +1,243 @@
...
+	path := s.HistoryPath()
+	data, err := os.ReadFile(path)
+	if errors.Is(err, os.ErrNotExist) {
+		return []*api.Message{}, nil
+	}
+	if err != nil {
+		return nil, err
+	}
+
+	var messages []*api.Message
```
droot: One thing I am wondering if we should move move to `jsonl` format where each line is just an individual json representing a message.

cmd/main.go
```diff
@@ -183,6 +184,7 @@ func (o *Options) InitDefaults() {
 	o.NewSession = false
 	o.ListSessions = false
 	o.DeleteSession = ""
+	o.SessionBackend = ""
```
droot: shouldn't we default to memory backend ?

cmd/main.go
```diff
@@ -399,54 +403,67 @@ func RunRootCommand(ctx context.Context, opt Options, args []string) error {
...
-	// TODO: Remove this when session persistence is default
 	if opt.NewSession || opt.ResumeSession != "" {
-		sessionManager, err = sessions.NewSessionManager()
+		sessionManager, err = sessions.NewSessionManager(opt.SessionBackend)
 		if err != nil {
 			return fmt.Errorf("failed to create session manager: %w", err)
 		}
 
-		// Handle session creation or loading
+		// Case 1: Create a completely new session
```
droot: This seems to be added by AI. We should avoid having "How" style comments in the code. Comments should mostly focus on the "Why" aspect here.

cmd/main.go
```diff
@@ -399,54 +403,67 @@ func RunRootCommand(ctx context.Context, opt Options, args []string) error {
...
 			if err != nil {
 				return fmt.Errorf("failed to create a new session: %w", err)
 			}
-			klog.Infof("Created new session: %s\n", chatStore.(*sessions.Session).ID)
+			klog.Infof("Created new session: %s\n", session.ID)
 		} else {
-			// Load existing session
-			var sessionID string
+			// Case 2: Resume an existing session
+			// Sub-case 2a: Resume the latest session (or default if no ID provided)
```
droot: See my earlier comment about the AI comments. I think you can instruct the AI to not do this.

ShubyM: I intentionally added the comments for the branches, because there was a lot of them in this case. I can remove them

pkg/agent/conversation.go
ShubyM: Initially I made the session reference private but making it public makes intialization easier

PR #668, opened 2026-05-21
fix(mcp): bind streamable-http listener to loopback by default

evilgensec:
## Summary

`cmd/mcp.go:178` formats the streamable-http listener endpoint as `:%d`, which causes Go's `net.Listen` to bind on all interfaces. The same code path registers the built-in `bash` and `kubectl` tools and dispatches calls to them with **no authentication, no Origin check, and no Host validation** (`handleBuiltinToolCall` invokes `tool.Run(ctx, args)` directly).

The documented setup in `docs/mcp-server.md:34` says *"This listens on http://localhost:9080/mcp by default"*, so users reasonably expect a loopback-only listener. In practice anyone on the same LAN (or the rest of a shared Cloud Build / coffee-shop / multi-tenant VM) can call the `bash` tool to run arbitrary shell as the running user, or the `kubectl` tool to read every Secret, ConfigMap and pod env in the user's configured clusters. A browser-driven DNS-rebinding attack can do the same against a single-user workstation that intends to use only loopback.

I verified locally on `v0.0.31` (`08cf256`):

```
$ kubectl-ai --mcp-server --mcp-server-mode streamable-http --http-port 19080
$ lsof -i :19080
COMMAND    PID USER FD TYPE DEVICE SIZE/OFF NODE NAME
kubectl-a 41136 end  7u IPv6 ...      0t0   TCP *:19080 (LISTEN)
```

Note `*:19080`, not `localhost:19080`. Unauthenticated `POST /mcp` with `{"jsonrpc":"2.0","method":"tools/call","params":{"name":"bash","arguments":{"command":"id; cat /etc/passwd"}}}` from any LAN host returns the command output.

## Fix

Introduce `--mcp-http-bind-address`, defaulting to `127.0.0.1`. The default forces a loopback bind which matches the documentation. Operators who intentionally want a remote listener can opt in with `--mcp-http-bind-address=0.0.0.0` (or any specific address); that mode emits a `klog.Warningf` so it is obvious in the build/CI logs that an unauth-able port is reachable from the network.

## Test plan

Local verification with the patched binary:

- [x] Default: `lsof -i :19183` shows `kubectl-a ... TCP localhost:19183 (LISTEN)` (loopback only).
- [x] `--mcp-http-bind-address 0.0.0.0` → `TCP *:19184 (LISTEN)` plus a warning log.
- [x] Build clean with `GOWORK=off go build -o /tmp/kubectl-ai-test ./cmd/`.

Future work happy to follow up with in a separate PR: optional bearer-token auth on the HTTP endpoint, Origin/Host allowlist for DNS-rebinding resistance, and routing built-in tool calls through the same approval/modifies_resource gate that protects the CLI/UI agent path.

### DIFF
--- a/cmd/main.go
+++ b/cmd/main.go
@@ -100,6 +100,10 @@ type Options struct {
 	MCPServerMode string `json:"mcpServerMode,omitempty"`
 	// Set the HTTP endpoint port for the MCP server when using HTTP transports like streamable-http.
 	HTTPPort int `json:"httpPort,omitempty"`
+	// MCPHTTPBindAddress is the host the MCP server listens on when using HTTP
+	// transports. Defaults to 127.0.0.1 because the MCP HTTP endpoint exposes the
+	// built-in `bash` and `kubectl` tools without authentication.
+	MCPHTTPBindAddress string `json:"mcpHTTPBindAddress,omitempty"`
 	// KubeConfigPath is the path to the kubeconfig file.
 	// If not provided, the default kubeconfig path will be used.
 	KubeConfigPath string `json:"kubeConfigPath,omitempty"`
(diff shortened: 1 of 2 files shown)


evilgensec:
Checking in on this. Anything I can do to help get it reviewed?

PR #669, opened 2026-05-22
fix: redact sensitive trace headers

### LINKED ISSUE #617: [Feature]: Trace file should filter sensitive headers
I've noticed that the generated trace.yaml includes all the headers for http requests, which can sometimes be sensitive (e.g,: goog-api-key when using gemini provider).

We should have some filtering for any provider that includes sensitive data, like google api key, openai api key, etc.

rohan-patnaik:
## What

Redacts sensitive HTTP headers before writing request and response entries to the trace journal.

The actual request sent to the provider is unchanged. Only the recorded trace output is sanitized.

## Why

Trace files are useful for debugging provider calls, but they can currently include headers such as `Authorization`, cookies, API keys, tokens, or secrets. That makes trace files risky to share or store.

Fixes #617.

## Testing

- `go test ./gollm -run TestJournalingRoundTripperRedactsSensitiveHeaders -count=1`
- `git diff --check`

Note: `go test ./gollm` currently fails on `TestGetAnthropicModel/explicit_model_is_highest_priority` on the current base branch. That failure is unrelated to this change.

### DIFF
--- a/gollm/http_journal.go
+++ b/gollm/http_journal.go
@@ -19,6 +19,7 @@ import (
 	"io"
 	"net/http"
 	"net/http/httputil"
+	"strings"
 
 	"github.com/GoogleCloudPlatform/kubectl-ai/pkg/journal"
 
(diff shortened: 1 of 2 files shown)


rohan-patnaik:
@noahlwest, would you have time to take a look when convenient? Thanks!

rohan-patnaik:
@noahlwest, could you take a look at this fix for #617 when convenient? Please let me know if anything else is needed from me. Thanks!

PR #670, opened 2026-05-22
feat: add suggest-only mode

rohan-patnaik:
## What

Adds a `--suggest-only` flag.

When enabled, kubectl-ai skips tool calls that would modify cluster resources. Instead of asking for approval or executing the command, it records the proposed action as skipped and sends that result back to the model so it can continue with a final suggestion.

Read-only tool calls can still run, so the agent can inspect the cluster before producing guidance.

## Why

Some users want kubectl-ai to help produce commands or manifests without making any changes to the cluster. Today they have to decline each proposed mutating command manually. This mode makes that workflow explicit and safer for production or GitOps-managed environments.

Fixes #628.

## Testing

- `go test ./pkg/agent -run 'TestAgentSuggestOnlySkipsMutatingTool|TestAgentEndToEndToolExecution' -count=1`
- `go test ./cmd ./pkg/agent -count=1`
- `git diff --check`

### DIFF
--- a/cmd/main.go
+++ b/cmd/main.go
@@ -84,6 +84,8 @@ type Options struct {
 	// SkipPermissions is a flag to skip asking for confirmation before executing kubectl commands
 	// that modifies resources in the cluster.
 	SkipPermissions bool `json:"skipPermissions,omitempty"`
+	// SuggestOnly skips execution of tool calls that would modify resources.
+	SuggestOnly bool `json:"suggestOnly,omitempty"`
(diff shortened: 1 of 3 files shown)


rohan-patnaik:
@droot, would you have time to review this when convenient? Thanks!

rohan-patnaik:
@droot, following up on your offer to review in #628—would you have a moment to look at this? Happy to make any adjustments needed. Thanks!

PR #683, opened 2026-08-06
feat(ollama): add session persistence

### LINKED ISSUE #464: [Feature]: Add session persistence for Ollama
#443 added session persistence for Gemini, but other providers were left unimplemented.

We should add similar implementation of Initialize() to load chat history for Ollama. This can happen in gollm/ollama.go and will likely follow a similar flow to Gemini, where we convert api messages to the provider-specific messages and add them to the history.

hugosmoreira:
## Summary

- restore persisted user, model, and agent messages into Ollama chat history
- preserve the configured Ollama system prompt while replacing stale conversation state
- skip unsupported or empty session records consistently with the other providers
- add regression coverage for role conversion, invalid records, repeated initialization, and session clearing

## Why

Ollama previously ignored persisted messages in `Initialize`, so loading or switching sessions could not restore the conversation. This converts stored API messages into Ollama messages while retaining the system prompt that Ollama keeps in its history.

## Validation

- `go test ./gollm -run '^TestOllama' -count=1`
- `go test -race ./gollm -run '^TestOllama' -count=1`
- `go test ./pkg/agent ./pkg/sessions -count=1`
- `go test ./... -count=1`
- `go test ./... -count=1` in `kubectl-utils`
- repository presubmits: build, vet, format, module tidiness, autogeneration, and mocks

Closes #464

### DIFF
--- a/gollm/ollama.go
+++ b/gollm/ollama.go
@@ -210,8 +210,56 @@ func (c *OllamaChat) SendStreaming(ctx context.Context, contents ...any) (ChatRe
 }
 
 func (c *OllamaChat) Initialize(messages []*kctlApi.Message) error {
-	klog.Warning("chat history persistence is not supported for provider 'ollama', using in-memory chat history")
+	history := make([]api.Message, 0, len(c.history)+len(messages))
+
+	// Ollama keeps the system prompt in its message history. Preserve it when
+	// replacing the conversation with a restored session or clearing a session.
+	for _, message := range c.history {
+		if message.Role == "system" {
+			history = append(history, message)
+		}
+	}
+
+	for _, message := range messages {
+		ollamaMessage, err := messageToOllama(message)
+		if err != nil {
+			continue
+		}
+		history = append(history, ollamaMessage)
+	}
+
+	c.history = history
 	return nil
+}
+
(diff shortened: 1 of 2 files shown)


hugosmoreira:
Quick update: I have a local, not-yet-pushed revision that preserves tool observations through Ollama session save/load. The new regression tests fail with the original implementation and pass with the fix, including checks of the actual outgoing request after filesystem save/load and clear. The focused tests also pass 20 repeated race-detector runs.

Full-suite verification still finds two failures that I also reproduced on the untouched PR head (`1fad35e`):

- `TestGetAnthropicModel/explicit_model_is_highest_priority`: expects `explicit-model`, but receives `env-model`.
- `TestAgentEndToEndToolExecution`: intermittently sees `running` where it expects `Done` or `WaitingForInput` (18 failures in 100 repetitions with `-race` on unchanged code).

These checks used Go 1.27.1 in WSL-backed Docker, not native Windows. I have left those unrelated tests and implementations unchanged.

Would you prefer these baseline failures handled in separate fixes before I mark this PR ready for review? I’m keeping this PR in draft and the revision local while those verification gates remain unresolved.

PR #688, opened 2026-09-04
Make resize-pvc eval portable across cluster types

### LINKED ISSUE #531: [Feature]: Make resize-pvc eval work well with any type of cluster
Right now, resize-pvc eval has a pretty low success rate on KinD clusters. Per https://github.com/GoogleCloudPlatform/kubectl-ai/pull/319#discussion_r2136601431, we would expect the LLMs to have some trouble getting this right since it's not very compatible with KinD clusters.

Ideally, we would have some way of testing LLMs ability to do this that is a bit more portable and will give us consistent results no matter what type of cluster we use.

Alternatively, we leave things as they are and leave this as a space for LLMs to grow. (apply the correct solution for different cluster types)

amarkdotdev:
## Summary
- Add a portable resize-pvc eval task override with a self-contained expandable StorageClass and static hostPath PV
- Verify the PVC requested size instead of waiting for PV capacity expansion, which is cluster-provisioner dependent
- Apply kubectl-ai eval task overrides when running dev/ci/periodics/run-evals.sh

Fixes #531

## Context
The upstream resize-pvc task relied on the default standard StorageClass and checked PersistentVolume capacity. That fails on KinD and other clusters where the default provisioner does not support volume expansion (see discussion in #319).

This override keeps the same user-facing prompt while making setup and verification portable.

## Test plan
- [ ] Run resize-pvc eval against a KinD cluster via run-evals.sh
- [ ] Confirm setup creates the expandable storage resources and the verifier passes after patching the PVC to 15Gi
- [ ] Confirm overrides are copied into the cloned k8s-ai-bench tasks directory before eval execution

### DIFF
--- a/dev/ci/periodics/run-evals.sh
+++ b/dev/ci/periodics/run-evals.sh
@@ -23,6 +23,13 @@ curl -sSL https://raw.githubusercontent.com/GoogleCloudPlatform/kubectl-ai/main/
 K8S_AI_BENCH_SRC="${REPO_ROOT}/.build/k8s-ai-bench-src"
 rm -rf "${K8S_AI_BENCH_SRC}"
 git clone https://github.com/gke-labs/k8s-ai-bench "${K8S_AI_BENCH_SRC}"
+
+EVAL_TASK_OVERRIDES_DIR="${REPO_ROOT}/eval-task-overrides"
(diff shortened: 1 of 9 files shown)


amarkdotdev:
any chance someone could take a first pass on the resize-pvc portability change? happy to tweak based on feedback.

